PR #154, opened 2015-03-04
Bugfix FlxScreenGrab.hx

JaytleBee:
Fixed a bug where the hotkey arguments would override grab()'s ones.
example:
FlxScreenGrab.defineHotKeys(["F12"], true);
FlxScreenGrab.grab(false);//will cause "save file" dialog to appear


Gama11:
Thanks!

PR #157, opened 2015-03-10
Fix FlxNestedSprite rendering with scale and angle, closes #156

### LINKED ISSUE #156: FlxNestedSprite doesn't seem to update scale and angle properly
Reproduction : 
1) Create a `parent` FlxNestedSprite and add 2 children
2) Set scale or angle to the parent. Or create an FlxTween on it's scale or angle

You can see that children are not following their relative x and y properly
scaled-up
angle-45deg
original

yanargyr:


yanargyr:
@Beeblerox Guys, why not merging yet? Do you have any concerns ? Is there something more I should do?

yanargyr:
Guys, any update on this pull? Is there some reason it's not getting merged?

I've added a demo here: https://drive.google.com/file/d/0BxA54yGcBV4UbGxQeThJU3dLQzQ/view?usp=sharing 

Please check it and give me some feedback

Gama11:
Sorry, seems like this has been forgotten.

Two things:

- seems to work well enough on flash, but the behavior on native targets / with tilesheet rendering is still broken?
- the demo is only compatible with the latest release / master, while this pull goes to the dev branch

flixel/addons/display/FlxNestedSprite.hx
```diff
@@ -266,14 +266,18 @@ class FlxNestedSprite extends FlxSprite
 					var cos:Float = Math.cos(radians);
 					var sin:Float = Math.sin(radians);
 					
-					var dx:Float = child.relativeX - offset.x;
-					var dy:Float = child.relativeY - offset.y;
+					/* Fix from yannicular */
```
Gama11: What's the point of this comment here? If you're concerned about credit - the commit history pretty much covers that.

We can't really put a comment in the source for the author of every fix, if we did readability would suffer. Additionally, only doing it in some cases would be unfair to other contributors.

yanargyr: On no, I'm sorry I forgot to remove this line, it was meant to help me while commenting out blocks of code while testing. I really didn't do that for the credit. Should I cancel this request, or post another  commit?
Thanx for your answer

Gama11: No worries. Adding commits to an existing pull request is fine.

yanargyr: About the `tilesheet rendering`, I'm not sure i got what you mean. Could you give me some more details ? 

On native targets I noticed there is still a glitch on neko, and it works ok with windows. Is there something else I'm missing ?

yanargyr: done

yanargyr:
on your first point, you mentioned ` tilesheet rendering` but I'm afraid I don't get what you mean? Can you please give me some more details ?

I just noticed there's still a problem with neko, bu it's fine on windows. Did you notice something else?

Gama11:
Tilesheet rendering is simply what's used for rendering on native targets.

I didn't even test windows, since the behavior of windows and neko is almost always identical when it comes to rendering. Quite surprising that that isn't the case here...

Gama11:
Btw, the demo project looks a lot better with `pixelPerfectRender = false` (no gaps / jittering between the sprites).

yanargyr:
Ok i got it now. I thought too that neko and windows should be similar and probably forgot to test neko thinking that testing on windows should be enough. Could it be that this difference is not related to this issue? 

I will upload a new project that works with the dev branch as soon as I can.

Btw, thanx for the tip on `pixelPerfectRender`. Do i need to set it for every individual LfxObject, or can I set it globally ?

btw, thanx for the tip on

Gama11:
I'm not really sure where the difference comes from, but it's probably unrelated to this, seeing how the behavior of the demo is different on neko vs windows without the fix as well.

Don't worry about dev compatiblity, I made it work with a few minor changes locally.

On the dev branch, there's `FlxG.camera.pixelPerfectRender` which works as a global setting.

yanargyr:
Thanx so much for your help @Gama11 It really means a lot since this was my first open-source contribution ever!

Gama11:
Windows and Neko *are* consistent actually, as long as the window dimensions are consistent. You had this line in the `Project.xml`:

`<window if="cpp" width="960" height="540" fps="60" background="#000000" hardware="true" vsync="true" />`

I opened a separate issue at #174.

PR #159, opened 2015-03-19
Fixes for openfl-next

larsiusprime:


Gama11:
I guess this makes it compile, but it doesn't work - `png` is null (lime 2.2.2).

PR #162, opened 2015-04-16
Tiled refactor

JoeCreates:
- Added a common base class for Tile and Object layers.
- Renamed Object groups to object layers to match the editor (the names have changed, anyway, due to the new structure, so it would be broken regardless).
- Getting a layer by name now uses a map of all layers instead of iterating over arrays.
- Maintains the order of all object and tile layers as they were in the editor (whereas before tile and object layers got separated, causing the order from the editor to be lost).

This will break the demos, of course.


Gama11:
>This will break the demos, of course.

I take it the pull to fix them is on the way? ;)

flixel/addons/editors/tiled/TiledObjectLayer.hx
```diff
@@ -0,0 +1,32 @@
...
+	public function new(source:Fast, parent:TiledMap)
+	{
+		super(source, parent);
+		type = TiledLayerType.OBJECT;
+		objects = new Array<TiledObject>();
+		color = source.has.color ? FlxColor.fromString(source.att.color) : FlxColor.TRANSPARENT;
+		loadObjects(source);
+	}
+	
+	private function loadObjects(source:Fast):Void {
```
Gama11: `{` should be in newline here.

Man, there really needs to be an auto-formatter for Haxe. :/

flixel/addons/editors/tiled/TiledMap.hx
```diff
@@ -135,73 +110,40 @@ class TiledMap
...
-			}
-		}
-		
-		return null;
+	public function getLayer(name:String):TiledLayer
+	{	
+		return layerMap.get(name);
 	}
 	
 	// works only after TiledTileSet has been initialized with an image...
```
Gama11: If you're already at it, you could change this to be a `/* */`-style comment. (FD ignores `//`-comments in completion).

flixel/addons/editors/tiled/TiledMap.hx
```diff
@@ -31,30 +31,29 @@ class TiledMap
 	
 	public var properties:TiledPropertySet;
 	
+	public var tilesets: Map<String, TiledTileSet>;
```
Gama11: There shouldn't be a space after `:` here.

flixel/addons/editors/tiled/TiledMap.hx
```diff
@@ -31,30 +31,29 @@ class TiledMap
...
+	private var layerMap:Map<String, TiledLayer>;
 	
-	// Use hash, we don't care about order
-	public var tilesets: Map<String, TiledTileSet>;
-	// Use array to preserve load order
-	public var layers:Array<TiledLayer>;
-	public var objectGroups:Array<TiledObjectGroup>;
-	
-	public function new(Data:Dynamic)
+	public function new(data:Dynamic)
```
Gama11: Not a fan of the `Dynamic` here, we should update this to use an asset abstract like core flixel does for images, sounds etc, something like:

`typedef FlxTiledAsset = OneOfTwo<String, Xml>;`

That way, we don't need the else-case that throws an error either.

I guess we would have the abstracts in `FlxAssets` non-private...

flixel/addons/editors/tiled/TiledTileLayer.hx
```diff
@@ -0,0 +1,222 @@
...
+				return tilesetID;
+			}
+		}
+		tiles.push(null);
+		return 0;
+	}
+
+	/**
+	 * Function that tries to resolve the tiles gid in the csv data.
+	 * TODO: It fails because I can't find a function to parse an unsigned int from a string :(
```
Gama11: Is this TODO still relevant?

flixel/addons/editors/tiled/TiledMap.hx
Gama11: They're available now: https://github.com/HaxeFlixel/flixel/commit/80a9d5a1a7270e32c35bc98c52ec71c41b8959e9

flixel/addons/editors/tiled/TiledMap.hx
JoeCreates: @Gama11 Yeah... sometimes miss them, but FD always adds them in. :P

flixel/addons/editors/tiled/TiledTileLayer.hx
JoeCreates: I couldn't tell.

flixel/addons/editors/tiled/TiledMap.hx
Gama11: It does? Never had that issue, what FD version are you running?

JoeCreates: It's with {} it adds the spaces. In a previous version I had a struct in there, and that's when the spaces were added.

JoeCreates:
@Gama11 I would suggest waiting to merge this until after LD, as I'll be using it for that and may discover issues or things that could be improved. Updates for the demos will have to wait until late next week.

flixel/addons/editors/tiled/TiledMap.hx
Gama11: Ah, I see. Yeah, structs aren't handled well...

Gama11:
Any update on this?

JoeCreates:
Sorry, been super busy lately. Will try to update this is a couple of days.

JoeCreates:
@Gama11 I've made the main changes and fixed the demos (https://github.com/HaxeFlixel/flixel-demos/pull/185)

Gama11:
Please update the changelog with a summary of the changes (especially any breaking ones). Not sure what the next flixel-addons release version is gonna be called, you can go with `?.?.?` or something liek that for now.

JoeCreates:
Done.

Gama11:
Thanks!

PR #163, opened 2015-04-21
FlxOutline now will not fill occupied pixels.

red-hara:


Gama11:
Got any concrete example where this fixes an issue?

Gama11:
No follow-up here.

Issue #161, opened 2015-03-25
About 'shapeDebug'...


xerosugar:
I would like to suggest that this is changed from private to public, or else one cannot use it to draw the debug lines, like in Nape's 'SpatialQueries' demo:
http://napephys.com/samples.html#swf-SpatialQueries

Issue #151, opened 2015-02-25
In FlxNapeSprite:addPremadeBody()


xerosugar:
If body is null and space haven't yet been set, this line will crash the program:
https://github.com/HaxeFlixel/flixel-addons/blob/dev/flixel/addons/nape/FlxNapeSprite.hx#L142

Maybe just change it to? -> NewBody.space = FlxNapeState.space;
Or change the method signature, letting the user provide a preferred space? -> addPremadeBody(NewBody:Body, ?_space:Space)

Edit: Btw, this happens when 'CreateRectangularBody' is set to false in the constructor.

Gama11:
Please always provide a reproducible code example in bug reports if possible. From your description it sounds like something like this would reproduce the error:

```haxe
var sprite = new FlxNapeSprite(0, 0, null, false);
sprite.addPremadeBody(new Body(BodyType.DYNAMIC, new Vec2(0, 0)));
```

(without a `FlxNapeSpace.init() call`)

which it doesn't. What am I missing?

xerosugar:
When I tried it a couple of minutes ago, it didn't crash, but the sprite's physics body wasn't there when I looked at the debug graphics. Did you check that as well? I'll try to provide code for issues in the future.

Gama11:
Yeah, but that's expected because the body I created in my snippet doesn't have a shape associated with it.

```haxe
FlxNapeSpace.init();
var sprite = new FlxNapeSprite(100, 100, null, false);
var body = new Body(BodyType.STATIC, new Vec2(0, 0));
body.shapes.add(new Polygon(Polygon.box(50, 50)));
sprite.addPremadeBody(body);
add(sprite);
```

You also need to set `physicsEnabled` to true explicitly.

xerosugar:
A small misunderstanding here I think :) I meant that, when I called addPremadeBody(bodyWithShapes) while having createRectangularBody = false, I was expecting it to work, since the comment for that parameter says "Whether to create a rectangular body for this sprite (use false if you want to add a custom body)." The code, with my change, feels more logical since you've made it clear in two places already 1) do not create a rectangularbody, and 2) addPremadeBody() that you, the user, want to use a custom-made Body for this sprite. What do you say?

Gama11:
Well, you did originally describe a bug / crash here, but instead it's unexpected behavior (`physicsEnabled` not being set to true)?

Gama11:
[...]

PR #165, opened 2015-05-11
fixes for issues #161 and #151

xerosugar:


flixel/addons/nape/FlxNapeSpace.hx
```diff
@@ -45,7 +45,7 @@ class FlxNapeSpace extends FlxBasic
 	public static var drawDebug(default, set):Bool;
 	
 	#if !FLX_NO_DEBUG
-	private static var shapeDebug:ShapeDebug;
+	public static var shapeDebug:ShapeDebug;
```
Gama11: Would be nice to have at least some basic documentation on what this is if it's made public.

flixel/addons/nape/FlxNapeSpace.hx
```diff
@@ -45,7 +45,15 @@ class FlxNapeSpace extends FlxBasic
 	public static var drawDebug(default, set):Bool;
 	
 	#if !FLX_NO_DEBUG
-	private static var shapeDebug:ShapeDebug;
+	/**
+	 * A useful "canvas" or sprite-like object which can be used to draw debug information on, such as lines, circles and more.
+	 * To give an even better idea of its use, see the official Nape demo 'SpatialQueries' (http://napephys.com/samples.html#swf-SpatialQueries)
+	 * where this is used to draw lines emitted from Rays.
+	 * To use this in HaxeFlixel, place the code that uses shapeDebug in your FlxState's draw()-method, but make sure it only runs when FlxNapeSpace.drawDebug = true, as it
```
Gama11: The single `=` is strange here in the context of programming, should either be `==` or just `is` IMO.

`as it will crash the program if it's not.` is not very specific, could just mention that it's null in that case.

xerosugar:
Changed it a bit, do you still feel that it's unclear?

Gama11:
That's better, though I still don't like "program will crash"-part.

I'll merge this manually since I have a different fix for #151 in mind if you're okay with that. Did you have a use case for providing a different `Space` instance for `addPremadeBody()` btw?

xerosugar:
How about now? Yes, of course, merge away :)
Nope, no use case other than for flexibility.

Gama11:
Already committed that here: https://github.com/HaxeFlixel/flixel-addons/commit/8f31de270e6faae25e03b4eb133129939792efaa

Well, yeah, but do you actually have a space other than `FlxNapeSpace.space`?

xerosugar:
Usually not, no... I guess I was thinking of the one user in the future that uses two Spaces, haha.

Gama11:
:) Using two spaces would actually be tricky with how `FlxNapeSpace` is implemented...

xerosugar:
tbh, I've never heard of someone using more than one space... :p Anyways, looking at the code now, and taking into account what you just said, I guess only this will do:

```if (currSpace == null) 
			{
				newBody.space = FlxNapeSpace.space;
			}
			else
			{
				newBody.space = currSpace;
			}
```

Gama11:
Any thoughts on the new behavior? faee801cbf3b9e1173d6d68e5d02e2b7035b9075

xerosugar:
Looks good to me, thanks!

PR #166, opened 2015-05-12
Fix FlxSpine when rendering using drawTriangles

Tiago-Ling:
FlxSpine depends on [spinehaxe](https://github.com/bendmorris/spinehaxe) which does not contain the property `RegionAttachment.uvtData`. This was probably a typo when updating the property names :)


Gama11:
I was wondering how this compiled in the first place...

```haxe
class RegionAttachment implements Dynamic<Dynamic>
```

Lovely, there go all the helpful compiler errors. :/

Anyway, thanks!

PR #172, opened 2015-06-18
Fixed problem with `TiledMap.layerMap`

SeiferTim:


Gama11:
Could you be a bit more specific than "problem"?

SeiferTim:
`layerMap` is used by `getLayer`, however, it was never being initialized or populated.

PR #173, opened 2015-06-18
Reconfigured Glitch, Wave, and Rainbow Sprite effects to no longer re…

SeiferTim:
…quire their targets to be added to a state and/or visible to work. Now supports chaining off of each other.


flixel/addons/effects/FlxGlitchSprite.hx
```diff
@@ -69,17 +86,12 @@ class FlxGlitchSprite extends FlxSprite
 		}
 		else
 		{
-			_time += elapsed;
+			_time += FlxG.elapsed;
```
Gama11: Hm, using FlxG.elapsed here is a pretty ugly hack that could cause unexpected sideeffects...

SeiferTim: elapsed only works if it comes through update, I was trying to avoid using update... any suggestions?

Gama11: It's somwhat tricky... if you call `target.update(elapsed)`, that results in two update calls per frame if both sprites are added.

SeiferTim: Yes, exactly why I wanted to avoid it... I don't know what else to change, though...

gamedevsam: You could implement the update function and cache lastElapsed...

gamedevsam: Please make that chance so we can merge this PR.

Gama11: This PR is kind of obsolete with https://github.com/HaxeFlixel/flixel-addons/pull/205.

gamedevsam:
I want to revisit this once we merge #205, I saw there was a new class FlxScrollZone, and potential improvements we might want for a 4.0 release.

Gama11:
Huh, `FlxScrollZone` was definitely not part of the original PR, looks like that got snuck in (accdientally?) in this commit: https://github.com/SeiferTim/flixel-addons/commit/30969868c2661c00ff2d53b8b6e5dc040661eba6

Either way, may as well close this, if any of this is going to be merged, it'll have to be in a new PR after #205 (without 15 merge commits).

PR #179, opened 2015-07-20
Added 'id' key for `tileProps`

SeiferTim:
This makes it easier to identify specific tiles when using the tilemap.


SeiferTim:
This was an error... committing the correct changes in another PR.

PR #180, opened 2015-07-20
Added 'id' key to `tileProps`

SeiferTim:
This makes it easier to find and identify tiles.


Gama11:
@JoeCreates Any thoughts?

JoeCreates:
In some cases, a user may have already added a custom id property to the tiles, which might have been an issue. The custom tile properties are added to the map afterwards, though, so the custom property will override this one. I think this is okay.

PR #183, opened 2015-08-25
Skip noLoadHash for Map Properties

gitfib:
this solves a bug when loading a map with Map Properties, because Tiled doesn't give it a name


Gama11:
Ping @JoeCreates 

(you're kinda the Tiled mantainer in my eyes now :P)

JoeCreates:
I think we should check if ```el.att.name```is null, instead of making a special case for ```properties```. After all, the thing that makes properties a special case is its lack of a name, and we can;t guarantee that other elements might also lack a name in the future.

PR #184, opened 2015-08-28
Fix transitions

larsiusprime:
Setting scroll factors to 0,0


larsiusprime:
Also fixed horizontal/vertical wipes to use a lot less memory.

larsiusprime:
(assume it's okay to unilaterally push a small change like this that passed travis?)

Gama11:
Sure.

flixel/addons/transition/TransitionFade.hx
```diff
@@ -154,22 +155,26 @@ class TransitionFade extends Transition
...
 			angle = DirY > 0 ? 90 : 270;
-			s.makeGraphic(FlxG.width, FlxG.height * 2, _data.color);
+			s.makeGraphic(1, FlxG.height * 2, _data.color);
 			pixels = s.pixels;
-			var gvert = FlxGradient.createGradientBitmapData(FlxG.width, FlxG.height, [_data.color, FlxColor.TRANSPARENT], 1, angle);
+			var gvert = FlxGradient.createGradientBitmapData(1, FlxG.height, [_data.color, FlxColor.TRANSPARENT], 1, angle);
 			pixels.copyPixels(gvert, gvert.rect, new Point(0, locY));
 			s.pixels = pixels;
+			s.scale.set(FlxG.width, 1.0);
+			s.updateHitbox();
```
Gama11: Do these sprites need hitboxes for anything?

larsiusprime:
Thx!

flixel/addons/transition/TransitionFade.hx
larsiusprime: Apparently yes. Without that, they don't render in the correct location.

madnotdead: I have an issue in current stable version when changing the hitbox of the sprite. When it renders the position depend on the edge of the hitbox.. i think is something related

PR #189, opened 2015-09-18
FlxTilemapExt with conditional tiles

adrianulima:
This modification allow the user to set some tiles to be conditionals.
Conditional tiles have collision, the user can set the condition to not
have collision.

`PlayState.hx`
`function create() {`
`level.setClouds([20], true);`
`level.setConditionals([21, 22, 23, 24]);`
`}`
`function update() {`
`level.throughDownClouds = FlxG.keys.anyPressed([DOWN, S]);`
`level.throughFloorSlopes = !FlxG.keys.anyPressed([UP, W]);`
`}`

I did also commit a [modified version of the demo](https://github.com/adrianulima/flixel-demos/commit/f73fa506d17216fe1a1838eb6b9922fdc74527bf).

### DIFF
--- a/flixel/addons/tile/FlxTilemapExt.hx
+++ b/flixel/addons/tile/FlxTilemapExt.hx
@@ -49,6 +49,12 @@ class FlxTilemapExt extends FlxTilemap
 	private var _slopeCeilLeft:Array<Int>;
 	private var _slopeCeilRight:Array<Int>;
 	
+	private var _slopeConditional:Array<Int>;
+	
+	public var throughDownClouds(default, set):Bool = false;
+	public var throughFloorSlopes(default, set):Bool = false;
+	public var throughCeilingSlopes(default, set):Bool = false;
+	
 	// Animated and flipped tiles related variables
 	private var _specialTiles:Array<FlxTileSpecial>;
 	
@@ -77,6 +83,7 @@ class FlxTilemapExt extends FlxTilemap
 		_slopeFloorRight = null;
 		_slopeCeilLeft = null;
 		_slopeCeilRight = null;
+		_slopeConditional = null;
 		
 		super.destroy();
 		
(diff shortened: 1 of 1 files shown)


flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -495,15 +502,49 @@ class FlxTilemapExt extends FlxTilemap
...
+	 */
+	public function setConditionals(?Slopes:Array<Int>):Void
+	{
+		if (Slopes != null)
+		{
+			_slopeConditional = Slopes;
+			
+			for (i in 0..._slopeConditional.length)
+			{
+				cast(_tileObjects[_slopeConditional[i]], FlxTile).allowCollisions = FlxObject.NONE;
```
Gama11: What's the deal with all these casts? They're completely unnecessary unless I'm missing something. The `Tile` type parameter is a `FlxTile` in `FlxTilemapExt`:

```haxe
$type(_tileObjects); // Array<flixel.tile.FlxTile>
```

flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -495,15 +502,49 @@ class FlxTilemapExt extends FlxTilemap
...
+	 * 
+	 * @param 	Slopes	An array containing the numbers of the slope tiles to be treated as conditional.
+	 */
+	public function setConditionals(?Slopes:Array<Int>):Void
+	{
+		if (Slopes != null)
+		{
+			_slopeConditional = Slopes;
+			
+			for (i in 0..._slopeConditional.length)
```
Gama11: Would be nicer if this was a for-each loop. Don't need the index.

flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -788,4 +896,19 @@ class FlxTilemapExt extends FlxTilemap
 		
 		return value;
 	}
+	
+	private function set_throughDownClouds(Value:Bool):Bool
```
Gama11: Why are these implemented as `(default, set)` properties? With these setters, a regular variable behaves the same.

flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -495,15 +502,49 @@ class FlxTilemapExt extends FlxTilemap
...
 	{
 		if (Clouds != null)
 		{
 			for (i in 0...(Clouds.length))
 			{
-				setTileProperties(Clouds[i], FlxObject.CEILING);			
+				setTileProperties(Clouds[i], FlxObject.CEILING);
+				if (Conditional)
+				{
+					cast(_tileObjects[Clouds[i]], FlxTile).callbackFunction = solveCollisionCloud;
```
Gama11: Using `callbackFunction` internally prevents the user from using it for anything else, which is not only inflexible, but also unexpected (there's nothing about `setClouds()` that hint at it behaving this way.

This could lead to some pretty hard-to-find bugs, like a `setClouds()` call overriding a user's custom callback.

flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -495,15 +502,49 @@ class FlxTilemapExt extends FlxTilemap
...
+		}
+	}
+	
+	/**
+	 * Solves collision against a cloud tile
+	 * 
+	 * @param 	Slope 	The slope to check against
+	 * @param 	Object 	The object that collides with the slope
+	 */
+	private function solveCollisionCloud(Slope:FlxObject, Object:FlxObject):Void
```
Gama11: This doesn't really "solve" any collisions, just set the `allowCollisions` property.

Gama11:
I'm not a huge fan of this API to be honest. The naming is confusing ("conditional" could mean a lot of things, you have to read the source code to understand that it conditionalizes _collisions_). It also has potential side-effects (overriding the `callbackFunction`).

It also doesn't really add any new functionality - you can do all the same things by using the existing `FlxTile#allowCollisions` bitfield (which is much more flexible). I think this might be a case of trying to do a little too much for the user. I guess it adds a bit of convenience for some use-cases, but it will just make things harder for others.

adrianulima:
I understand you, and agree. Thanks for this complete feedback. 👍

I'm beginner with Haxe and this is very good to me.

Can I try the other modification that implement others degrees slopes (67.5 and 22.5 and maybe a custom degree)?

And I also could make this conditional tiles modification in the demo project without change the FlxTilemapExt, what you think?

Gama11:
Sure, more slope types sounds like a useful addition (although at some point, it would probably be easier to just use Nape for collisions with complex terrain :) ).

Implementing it in the demo as an example of how something like that can be done seems sensible - although it might be weird to have that in the FlxTilemapExt demo since it's not really specfic to that (can be done with a regular FlxTilemap).
Another option would be to add some examples for tilemap collisions / allowCollisions etc to the haxeflixel mechanics site (https://github.com/HaxeFlixel/haxeflixel-mechanics).

PR #190, opened 2015-09-22
Change FlxScreenGrab from systools to linc_dialogs

larsiusprime:


Gama11:
I can't compile this - get this error for some reason:

```haxe
C:/HaxeToolkit/haxe/lib/flixel-addons/git/flixel/system/macros/FlxConditionals.hx:117: characters 69-88 : FLX_MOUSE_ADVANCED can only be defined by flixel.
Build halted with errors.
```

Doesn't really make too much sense...

larsiusprime:
Do you have linc_dialogs?

Gama11:
I do, but it doesn't seem to make any difference whether I include it or not.

larsiusprime:
Yeah not sure what's going on. Make sure your libraries are updated I guess? I dont' get anything like that here.

Gama11:
Would be nice if somebody else could test this to see whether it's an issue on my side or not.

larsiusprime:
Travis isn't complaining, are you still having this problem @Gama11 ? @Tiago-Ling  and @ProG4mr and @gamedevsam , does this work for you guys or not?

nazywam:
Works fine for me on both Windows and Mac

adrianulima:
Tested on Windows, worked fine.

Gama11:
For whatever reason, my original issue is gone and I can compile this now. However, I noticed some other issues:

- no file dialog appears on neko
- on cpp, the screenshot is saved file, but immediately after clicking "ok" in the file dialog the game crashes

This was tested in Mode with the following code added to `PlayState#update()`:

```haxe
if (FlxG.keys.justReleased.I)
	FlxScreenGrab.grab(null, true);
```

larsiusprime:
I'll test and see what I can do about these. It may well be that linc_dialogs does not work in neko, in which case systools will still be required for that target. 

If that is the case (and presuming I fix the other issue), then this would still be an improvement as it would allow for a non-LGPL licensed library for C++ target (neko mostly being for testing).

larsiusprime:
I'm testing this now, sorry the commit history has gotten so messy, might have to restart it XD

larsiusprime:
> no file dialog appears on neko

I get this behavior too. I guess linc_dialogs is not supported on neko?

> on cpp, the screenshot is saved file, but immediately after clicking "ok" in the file dialog the game crashes

The game does not crash for me in cpp (windows), but it does make a small error -- if the user doesn't manually add ".png" it saves the file with no extension. Quick line or two fixes that.

-----------------

In light of these findings, what's the recommended path forward? Systools will prevent people from statically linking their games as it's LGPL, and using linc_dialogs will allow windows/mac/linux builds to not worry about that, which I imagine are the chief candidates for static linking.

My proposed solution:

1) support both systools and linc_dialogs
2) if linc_dialogs is present, and target is not flash or neko -- use linc_dialogs, otherwise check:
3) if systools is present, and sys is defined -- use systools
4) otherwise throw an error that the relevant library is not present.

larsiusprime:
@gama11 just tested flxscreengrab, ready for review? Thanks for your patience.

Gama11:
That solution sounds reasonable, however it does make me wonder...

- is it even worth supporting file dialogs with how much of a hassle it is cross-platform? That's 3 different implementations for flash, neko and cpp. Could just let the user deal with it?
- if we do decide to support it, is `FlxScreenGrab` still the right place for it with all the complexity? Do we need a separate wrapper ("`FlxSaveDialog`")?

larsiusprime:
I do think a FlxSaveDialog wrapper makes sense as we might have use for it in other places and flixel users will appreciate not having the hassle.

FlxScreenGrab seems like a useful enough utility to go through this minor bit of pain for it.

larsiusprime:
Obsolete, see https://github.com/HaxeFlixel/flixel-addons/pull/207

Issue #113, opened 2014-09-23
No vertical mode in FlxWaveSprite


dwarvenbuffet:
As of right now there seems to be no way to apply a vertical wave effect to a sprite despite FlxGlitchSprite supporting directions.

PR #191, opened 2015-09-24
FlxTilemapExt with more slope degrees, closes #188

### LINKED ISSUE #188: FlxTilemapExt more Extended
Hi, I'm new with HaxeFlixel and with Git repositories. Last year I used HaxeFlixel in some tests, but since then I haven't used it. This year I worked just with `AS3`, `Objective C` and `Swift`.
Now I decided back to `Haxe` learning and, if I can, contribute with the community.

FlxTilemapExt

Last year I made some modifications to `FlxTilemapExt` to allow player down in cloud tiles. I also added two more levels of slopes, **67.5** and **22.5** degrees.
As were my first code with HaxeFlixel, they are not optimized and are not compatible with the version **4.0.0**.

Slopes Example
*My [Twitter](https://twitter.com/adrianulima) has more images.*

##### So, could I try rewrite this modifications as improvements to this repository?
- ~~Possibility to walk through slope tiles~~
- [x] Possibility to player fall in cloud tiles
- [x] 67.5° Slopes floor and ceiling
- [x] 22.5° Slopes floor and ceiling
- [x] Jump walls
- [x] Reversal of gravity

adrianulima:
FlxTilemap with slopes of 22.5 and 67.5 degrees, floor and ceiling.
I also edited some parts of the code that I think was unnecessary

You can try a [swf demo here](https://drive.google.com/file/d/0B5-AbrEtUqC8dDlkemlJdExoVFk/view?usp=sharing)

This demo is a modified version of [FlxTilemapExt platformer demo](https://github.com/adrianulima/flixel-demos/tree/dev/Platformers/FlxTilemapExt) with jump walls e gravity inversion to test ceiling slopes.

### DIFF
--- a/flixel/addons/tile/FlxTilemapExt.hx
+++ b/flixel/addons/tile/FlxTilemapExt.hx
@@ -1,24 +1,20 @@
 package flixel.addons.tile;
 
 import flash.display.BitmapData;
-import flash.geom.Matrix;
-import flash.geom.Point;
-import flash.geom.Rectangle;
 import flixel.addons.tile.FlxTilemapExt;
 import flixel.addons.tile.FlxTileSpecial;
 import flixel.FlxCamera;
 import flixel.FlxG;
 import flixel.FlxObject;
 import flixel.graphics.frames.FlxFrame;
-import flixel.graphics.tile.FlxDrawTilesItem;
+import flixel.graphics.frames.FlxFramesCollection;
+import flixel.math.FlxMath;
 import flixel.math.FlxMatrix;
+import flixel.math.FlxPoint;
 import flixel.tile.FlxTile;
 import flixel.tile.FlxTilemap;
 import flixel.tile.FlxTilemapBuffer;
 import flixel.util.FlxDestroyUtil;
-import flixel.math.FlxMath;
-import flixel.math.FlxPoint;
-import flixel.graphics.frames.FlxFramesCollection;
 
 // TODO: add support for tilemap scaling
 // TODO: try to make it cleaner (i mean rendering and animated tiles)
@@ -293,14 +289,14 @@ class FlxTilemapExt extends FlxTilemap
 		for (i in 0...tiles.length) 
 		{
 			tile = tiles[i];
-			if (t != null && t.isSpecial())
+			if (tile != null && tile.isSpecial())
 			{
 				_specialTiles[i] = tile;
 				
 				tile.currTileId -= _startingIndex;
 				tile.frames = this.frames;
 				
-				if (t.hasAnimation()) 
+				if (tile.hasAnimation()) 
 				{
 					var animFrames:Array<Int> = tile.animation.frames;
 					var preparedFrames:Array<Int> = [];
--- a/flixel/addons/tile/FlxTilemapExt.hx
+++ b/flixel/addons/tile/FlxTilemapExt.hx
@@ -34,11 +34,6 @@ import flixel.graphics.frames.FlxFramesCollection;
  */
 class FlxTilemapExt extends FlxTilemap
 {
-	public static inline var SLOPE_FLOOR_LEFT:Int = 0;
-	public static inline var SLOPE_FLOOR_RIGHT:Int = 1;
-	public static inline var SLOPE_CEIL_LEFT:Int = 2;
-	public static inline var SLOPE_CEIL_RIGHT:Int = 3;
-	
 	// Slope related variables
 	private var _snapping:Int = 2;
 	private var _slopePoint:FlxPoint;
@@ -49,11 +44,10 @@ class FlxTilemapExt extends FlxTilemap
 	private var _slopeCeilLeft:Array<Int>;
 	private var _slopeCeilRight:Array<Int>;
 	
-	private var _slopeConditional:Array<Int>;
-	
-	public var throughDownClouds(default, set):Bool = false;
-	public var throughFloorSlopes(default, set):Bool = false;
-	public var throughCeilingSlopes(default, set):Bool = false;
+	private var _slope22High:Array<Int>;
+	private var _slope22Low:Array<Int>;
+	private var _slope67High:Array<Int>;
+	private var _slope67Low:Array<Int>;
 	
 	// Animated and flipped tiles related variables
 	private var _specialTiles:Array<FlxTileSpecial>;
@@ -70,6 +64,11 @@ class FlxTilemapExt extends FlxTilemap
 		_slopeCeilLeft = new Array<Int>();
 		_slopeCeilRight = new Array<Int>();
 		
+		_slope22High = new Array<Int>();
+		_slope22Low = new Array<Int>();
+		_slope67High = new Array<Int>();
+		_slope67Low = new Array<Int>();
+		
 		// Flipped/rotated tiles variables
 		_specialTiles = null;
 	}
@@ -83,7 +82,11 @@ class FlxTilemapExt extends FlxTilemap
 		_slopeFloorRight = null;
 		_slopeCeilLeft = null;
 		_slopeCeilRight = null;
-		_slopeConditional = null;
+		
+		_slope22High = null;
+		_slope22Low = null;
+		_slope67High = null;
+		_slope67Low = null;
 		
 		super.destroy();
 		
@@ -467,12 +470,6 @@ class FlxTilemapExt extends FlxTilemap
 						overlapFound = (Object.x + Object.width > tile.x) && (Object.x < tile.x + tile.width) && (Object.y + Object.height > tile.y) && (Object.y < tile.y + tile.height);
 					}
 					
-					// Solve slope collisions if no overlap was found
-					/*
-					if (overlapFound
-						|| (!overlapFound && (tile.index == SLOPE_FLOOR_LEFT || tile.index == SLOPE_FLOOR_RIGHT || tile.index == SLOPE_CEIL_LEFT || tile.index == SLOPE_CEIL_RIGHT)))
-					*/
-					
 					// New generalized slope collisions
 					if (overlapFound || (!overlapFound && checkArrays(tile.index)))
 					{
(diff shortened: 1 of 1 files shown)


flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -713,21 +890,21 @@ class FlxTilemapExt extends FlxTilemap
 	 */
 	private function setSlopeProperties():Void
 	{
-		for (i in 0..._slopeFloorLeft.length)
+		for (i in _slopeFloorLeft)
```
Gama11: I like the change to for each loops, but `i` doesn't really make sense anymore as a variable name in that case, since it generally stands for "index". Should be `tile` or `slope`.

Gama11:
Gonna need a pull request for flixel-demos as well. :)

flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -539,6 +524,79 @@ class FlxTilemapExt extends FlxTilemap
 	}
 	
 	/**
+	 * Sets the slope arrays, which define which tiles are treated as slopes of 22.5 degrees.
+	 * 
+	 * @param 	HighSlopes 	An array containing the numbers of the tiles to be treated as high slope tiles with a slope on the left.
+	 * @param 	LowSlopes	An array containing the numbers of the tiles to be treated as low slope tiles with a slope on the right.
+	 */
+	public function setSlopes22(?HighSlopes:Array<Int>, ?LowSlopes:Array<Int>)
```
adrianulima: The user must set the slops of 22.5 and 67.5 degrees in the `setSlopes` method as the default ones.

But he need to identify if it is a 22 of 67 and if is a High or a Low slope.
I think the names of functions and of params are not yet intuitive enough.
slopenames

Need suggestions here...

Gama11: "steep" and "flat" might be an option?

larsiusprime: How about "steep" and "gentle" ?

adrianulima: "steep" and "gentle" are valid function names:
`setSteep(HighSlopes, LowSlopes)` / `setGentle(HighSlopes, LowSlopes)`
or
`setSteep(ThickSlopes, ThinSlopes)` / `setGentle(ThickSlopes, ThinSlopes)`

larsiusprime: Here's another thought. It looks kind of like a foot, right?

"Heel" and "Toe"

adrianulima: It's your decision, guys..
`setSteep(HeelSlopes, ToeSlopes)` / `setGentle(HeelSlopes, ToeSlopes)` ?
I have to do something more here?

And I have another modified class, a FlxWaveSprite with vertical mode that fix https://github.com/HaxeFlixel/flixel-addons/issues/113
To pull-request it, do I have to make another fork? or another branch? or just pull-request together here?

Gama11: Another branch.

Gama11:
Even with better method names, I don't think the API is gonna be very intuitive to use. How about something like this?

```haxe
setSlopes(tiles:Array<Int>, steepness:FlxSlopeSteepness, flipX:Bool = false, flipY:Bool = false)
```

```haxe
FlxSlopeSteepness
 - GENTLE // 22.5°
 - MODERATE // 45°
 - STEEP // 67.5°
```

Those 3 parameters are enough to determine the specific slope type, and you only need a single function.

Flipping would be mapped out like this:

| Slope | `flipX` | `flipY` |
|-------|---------|---------|
| ◢     |   `false`  |   `false`      |
| ◣     |    `true`  | `false`        |
|  ◥     |  `false`       | `true`        |
|  ◤     |  `true`       | `true`        |

adrianulima:
@Gama11 I think a perfect idea to use Enums to set the steepness. But I think `FlipX` and `FlipY` will confuse with flipped tiles, user will think that the class automatically flip tiles or something like that.

And we also still need to set which tiles are `THICK` and `THIN`

We could use Enums for Slopes too like:
```haxe
FlxSlopeDirection
 - NORTHEAST // ◣
 - SOUTHEAST // ◤
 - SOUTHWEST // ◥
 - NORTHWEST // ◢
```
```haxe
//this automatically set slopes `THIN`
setSlopes(tiles:Array<Int>, direction:FlxSlopeDirection, steepness:FlxSlopeSteepness = MODERATE)
//and user need to set `THICK`
setThickSlopes(tiles:Array<Int>)

//-------OR-------
setSlopes(tiles:Array<Int>, direction:FlxSlopeDirection, steepness:FlxSlopeSteepness = MODERATE, ?tickTiles:Array<Int>)
```
what you think?

Gama11:
That's a good point about flipX / Y - especally since `FlxTileSpecial` has properties with those exact name that do just that (flip the graphic). I like the `FlxSlopeDirection` idea.

I'm not sure what you mean by thick and thin though.

adrianulima:
We are setting `steepness` and `direction`, but we still need to identify which are thicker
slopes

Gama11:
I don't understand - for the left tile, you'd choose `FlxSlopeDirection.NORTHWEST` and `FlxSlopeSteepness.STEEP`, `FlxSlopeDirection.NORTHEAST` and `FlxSlopeSteepness.GENTLE` for the right one - what information is missing there?

adrianulima:
It's because in this image have 4 tiles.. one `HIGH` and one `LOW` tiles for each slope.
Red lines are tile spaces, you see?

Gama11:
Oh, yeah, I see it now. :) Hm, not sure what a good API for this would be... Maybe just split the gentle and steep values of the enum into two?

```haxe
FlxSlopeSteepness
 - GENTLE_LOWER
 - GENTLE_UPPER
 - MODERATE
 - STEEP_LOWER
 - STEEP_UPPER
```

adrianulima:
I think this way the user would have to use a lot of `setSlopes()` to set all tiles, more than 12 lines

Maybe the solution is back to initial idea:
```haxe
//set the direction of all slopes as original function is
setSlopes(?northeast:Array<Int>, ?northwest:Array<Int>, ?southEast:Array<Int>, ?southWest:Array<Int>);

//identify which are steep / gentle, and also which are thin / thick
setSteepness(steepness:FlxSlopeSteepness, thinTiles:Array<Int>, ?thickTiles:Array<Int>);

FlxSlopeSteepness
 - GENTLE // 22.5°
 - MODERATE // 45°
 - STEEP // 67.5°
```

I think this two functions would solve everything with one call of `setSlopes()` and two of `setSteepnes()`

flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -777,15 +910,22 @@ class FlxTilemapExt extends FlxTilemap
...
 			}
 		}
 		
 		return value;
 	}
+}
+
+enum FlxSlopeSteepness
+{
+	GENTLE; // 22.5°
```
Gama11: This type of comment doesn't work properly for completion, it has to be a `/** */`-style comment _before_ the actual enum value - see `FlxCameraFollowStyle` for an example.

flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -493,49 +494,110 @@ class FlxTilemapExt extends FlxTilemap
...
 	 * 
-	 * @param 	LeftFloorSlopes 	An array containing the numbers of the tiles to be treated as floor tiles with a slope on the left.
-	 * @param 	RightFloorSlopes	An array containing the numbers of the tiles to be treated as floor tiles with a slope on the right.
-	 * @param 	LeftCeilSlopes		An array containing the numbers of the tiles to be treated as ceiling tiles with a slope on the left.
-	 * @param 	RightCeilSlopes		An array containing the numbers of the tiles to be treated as ceiling tiles with a slope on the right.
+	 * @param 	ThickTiles 	An array containing the numbers of the tiles to be treated as thick slope.
+	 * @param 	ThinTiles	An array containing the numbers of the tiles to be treated as thin slope.
 	 */
-	public function setSlopes(?LeftFloorSlopes:Array<Int>, ?RightFloorSlopes:Array<Int>, ?LeftCeilSlopes:Array<Int>, ?RightCeilSlopes:Array<Int>):Void
+	public function setSteepness(steepness:FlxSlopeSteepness, ThickTiles:Array<Int>, ?ThinTiles:Array<Int>)
```
Gama11: Is there any reason to call this function with `FlxSlopeSteepness.GENTLE` if that's the default? If that wasn't possible, `ThinTiles` wouldn't have to be optional (which seems a little weird).

Also, `setSlopeSteepness()` might be a better name?

adrianulima: The default is `FlxSlopeSteepness.MODERATE` but I know what you mean. The reason was to permit the user set the slope back to moderate, and in this case, he would use just `ThickTiles`.
I think slopes will never change angles, so this really is unnecessary.

So I think we don't need the `FlxSlopeSteepness` Enum, and can back to `setGentle()` and `setSteep()` separeted functions

what you think?
`setGentle(ThickTiles:Array<Int>, ThinTiles:Array<Int>)`
`setSteep(ThickTiles:Array<Int>, ThinTiles:Array<Int>)`

flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -139,22 +146,8 @@ class FlxTilemapExt extends FlxTilemap
...
-		if (screenYInTiles < 0)
-		{
-			screenYInTiles = 0;
-		}
-		if (screenYInTiles > heightInTiles - screenRows)
-		{
-			screenYInTiles = heightInTiles - screenRows;
-		}
+		screenXInTiles = Std.int(FlxMath.bound(screenXInTiles, 0, widthInTiles - screenColumns));
+		screenYInTiles =  Std.int(FlxMath.bound(screenYInTiles, 0, heightInTiles - screenRows));
```
adrianulima: I changed lines here and in `397`, is it ok?

flixel/addons/tile/FlxTilemapExt.hx
Gama11: I guess that's ok too, the names should be a bit more specific though: `setGentleSlopes()` and `setSteepSlopes()`.

Still not sure if "thin" / "think" is a good description for the upper and lower part of a slope.

flixel/addons/tile/FlxTilemapExt.hx
Gama11: Nice cleanup - 15 lines of code that now only need 2. :+1:

Pretty sure the logic is still the same.

flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -311,7 +304,7 @@ class FlxTilemapExt extends FlxTilemap
 	
 	/**
 	 * THIS IS A COPY FROM FlxTilemap
-	 * I've only swapped lines 386 and 387 to give DrawTilemap() a chance to set the buffer dirty
```
Gama11: Hm, yeah, referencing the exact lines here is a bad idea since that's bound to change...

Maybe there should be a comment directly before the mentioned change.

adrianulima: Actually I didn't dig in this method, but I think is a @mrcdk modification. Comparing to original `FlxTilemap` seems a little outdated. I don't know how `FlxTileSpecial` works neither how to use it, may be good someone else check it.

Gama11: I know, but you changed the comment, which is why it caught my attention. :)

Gama11:
Thanks, I think this turned out really nicely!

PR #192, opened 2015-09-25
Vertical mode in FlxWaveSprite, closes #113

### LINKED ISSUE #113: No vertical mode in FlxWaveSprite

adrianulima:
add vertical direction and fix a problem with `FlxWaveMode.TOP` that `_time`
was always `= 0`

Fix issue https://github.com/HaxeFlixel/flixel-addons/issues/113

### DIFF
--- a/flixel/addons/effects/FlxWaveSprite.hx
+++ b/flixel/addons/effects/FlxWaveSprite.hx
@@ -136,15 +136,15 @@ class FlxWaveSprite extends FlxSprite
 	{
 		var oldGraphic:FlxGraphic = graphic;
 		
-		var horizontalStrength = (direction == HORIZONTAL) ? strength * 2 : 0;
-		var verticalStrength = (direction == VERTICAL) ? strength * 2 : 0;
+		var horizontalStrength = (direction == HORIZONTAL) ? strength : 0;
+		var verticalStrength = (direction == VERTICAL) ? strength : 0;
 		target.drawFrame(true);
-		setPosition(target.x - strength, target.y);
+		setPosition(target.x - horizontalStrength, target.y - verticalStrength);
 		makeGraphic(
 			Std.int(target.frameWidth + horizontalStrength * 2),
 			Std.int(target.frameHeight + verticalStrength * 2),
 			FlxColor.TRANSPARENT, true);
-		_flashPoint.setTo(strength, 0);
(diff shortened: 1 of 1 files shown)


flixel/addons/effects/FlxWaveSprite.hx
```diff
@@ -98,26 +103,25 @@ class FlxWaveSprite extends FlxSprite
...
-		pixels.unlock();
 		
-		if (_targetOffset == -999)
-		{
-			_targetOffset = offset;
-		}
-		else
-		{
-			if (offset == _targetOffset)
-				_time = 0;
```
adrianulima: I did'nt find a use of `_targetOffset`, and whe using `FlxWaveMode.TOP` the wave wasn't updating because this `_time = 0`

Gama11:
Nice!

The demo needs to be updated for this as well.

I guess it's a matter of perspective, but the directions seem reversed to me - I'd consider the previous direction to be "vertical", while what you added makes the waves go in horizontal direction.

adrianulima:
I can update the demo tomorrow and I will reverse direction names.
What you think about `TOP` and `BOTTOM` names?

adrianulima:
I realized that swapping the direction names it becomes inconsistent with FlxGlitchSprite.

Should I back to previous names?
And about `TOP` and `BOTTOM` names, could be `BEGIN` and END`?

Gama11:
How does it become inconsistent? Guess I'm gonna have to take another look.

I was never a big fan of those wave modes to begin with - they seem quite arbitrary and inflexible. It might be better if you could define the offset yourself?

adrianulima:
>"It might be better if you could define the offset yourself?"

This `wavelength` parameter solves?
I also updated the demo https://github.com/HaxeFlixel/flixel-demos/pull/203

And I mean `FlxWaveSprite` HORIZONTAL direction is more like VERTICAL in `FlxGlitchSprite` now

Gama11:
Yeah, I guess you're right about the directions...

The `wavelength` option seems quite useful! :+1:

`TOP` should become `START` (rather than `BEGIN`).

Gama11:
Thanks!

Fyi, GitHub doesn't send notifications for new commits on pull requests, so a commit when you made a change doesn't hurt. :)

PR #193, opened 2015-09-29
This fix the unused dragging callbacks, closes #95

### LINKED ISSUE #95: Unused dragging callbacks in FlxExtendedSprite
Inside FlxExtendedSprite.hx, `mouseStartDragCallback` and `mouseStopDragCallback` do not fire. They can be set, but are never read except to be set to null in `disableMouseDrag()`.

adrianulima:
Now fires `mouseStartDragCallback` and `mouseStopDragCallback`
Fix issue: https://github.com/HaxeFlixel/flixel-addons/issues/95

Why this class is `FlxExtendedSprite` and the extended version of `FlxTilemap` is called `FlxTilemapExt`?

Maybe we should rename this class to `FlxSpriteExt` to be more clean, short and similar to `FlxTilemapExt`, what you think?

### DIFF
--- a/flixel/addons/display/FlxExtendedSprite.hx
+++ b/flixel/addons/display/FlxExtendedSprite.hx
@@ -674,6 +674,11 @@ class FlxExtendedSprite extends FlxSprite
 			_dragOffsetX = Std.int(frameWidth / 2);
 			_dragOffsetY = Std.int(frameHeight / 2);
 		}
+		
+		if (mouseStartDragCallback != null)
+		{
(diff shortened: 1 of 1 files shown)


Gama11:
Thanks!

They are named differently because they have different authors.

Not sure if renaming is worth it, it's a breaking change after all. Plus, most of the docs you can find on former Flixel Power Tools classes like FlxExtendedSprite are only available in AS3 / the power tools demo. Those resources won't be renamed, which would cause confusion.

PR #194, opened 2015-09-30
FlxTypeText prevent word jumping to next line, fixes #40

### LINKED ISSUE #40: FlxTypeText: find a way to avoid "jumping" with word-wrap by checking if there is space beforehand
https://github.com/HaxeFlixel/flixel-demos/issues/63

adrianulima:
Added a function that replace last space in a line for a line break. To
prevent a word start typing in a line and jump to next.
created property `preventWordJump` to set it.

solves issue: https://github.com/HaxeFlixel/flixel-addons/issues/40

### DIFF
--- a/flixel/addons/text/FlxTypeText.hx
+++ b/flixel/addons/text/FlxTypeText.hx
@@ -44,6 +44,10 @@ class FlxTypeText extends FlxText
 	 */
 	public var cursorCharacter:String = "|";
 	/**
+	 * Set to true to prevent a word start typing in a line and jump to next line.
+	 */
+	public var preventWordJump:Bool = true;
+	/**
 	 * The speed at which the cursor should blink, if shown at all.
 	 */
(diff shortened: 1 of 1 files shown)


flixel/addons/text/FlxTypeText.hx
```diff
@@ -44,6 +44,10 @@ class FlxTypeText extends FlxText
 	 */
 	public var cursorCharacter:String = "|";
 	/**
+	 * Set to true to prevent a word start typing in a line and jump to next line.
+	 */
+	public var preventWordJump:Bool = true;
```
Gama11: More of a line than a word jump I guess?

Any reason you'd wanna disable this?

adrianulima:
Changed the name to preventLineJump. I didn't realize performance difference, but I left the setter to have more flexibility.
may I remove it?

Gama11:
I removed the option - don't think there's really a use case for it.

PR #195, opened 2015-10-01
Added FlxPieDial

larsiusprime:
Quick demo code:

```haxe
  var pieDial = new FlxPieDial(100, 100, 10, FlxColor.LIME, 72, FlxPieDial.CIRCLE);
  add(pieDial);
  pieDial.amount = 0;
  FlxTween.tween(pieDial, { amount:1.0 }, 2.0, {type:FlxTween.PINGPONG} );
```

### DIFF
--- a/flixel/addons/display/FlxPieDial.hx
+++ b/flixel/addons/display/FlxPieDial.hx
@@ -1,0 +1,236 @@
+package flixel.addons.display;
+
+import flixel.FlxSprite;
+import flixel.math.FlxPoint;
+import flixel.math.FlxVector;
+import flixel.util.FlxColor;
+import flixel.util.FlxSpriteUtil;
+import haxe.EnumTools;
+import openfl.display.BitmapData;
+import openfl.display.BitmapDataChannel;
+import openfl.display.BlendMode;
+
+/**
+ * A dynamic shape that fills up the way a pie chart does. Useful for timers and other things.
+ * @author larsiusprime
+ */
+class FlxPieDial extends FlxSprite
+{
+
+	public var amount(default, set):Float = 0;
+	private var pieFrames:Int = 0;
+	
+	public static inline var CIRCLE:String = "circle";
+	public static inline var SQUARE:String = "square";
+	
+	public function new(X:Float, Y:Float, Radius:Int, Color:FlxColor=FlxColor.WHITE, Frames:Int=36, Shape:String = CIRCLE, Clockwise:Bool = true) 
+	{
+		super(0, 0);
+		makePieDialGraphic(Radius, Color, Frames, Shape, Clockwise);
+	}
+	
+	public function set_amount(f:Float):Float
+	{
+		amount = f;
+		var frame:Int = Std.int(f * pieFrames);
+		animation.frameIndex = frame;
+		return amount;
+	}
+	
+	private function makePieDialGraphic(Radius:Int, Color:FlxColor, Frames:Int = 36, Shape:String, Clockwise:Bool)
+	{
+		pieFrames = Frames;
+		var key:String = "pie_dial_" + Color.toHexString() + "_" + Std.string(Radius) + "_" + Frames + "_" + Shape;
+		var W = Radius * 2;
+		var H = Radius * 2;
+		if (false == FlxG.bitmap.checkCache(key))
+		{
+			var bmp = makePieDialGraphicSub(Radius, Color, Frames, Shape, Clockwise);
+			FlxG.bitmap.add(bmp, true, key);
+		}
+		
+		loadGraphic(key, true, W, H);
+		
+		amount = 1;
+	}
+	
+	private function makePieDialGraphicSub(Radius:Int, Color:Int, Frames:Int, Shape:String, Clockwise:Bool):BitmapData
+	{
+		var W = Radius * 2;
+		var H = Radius * 2;
+		
+		var rows:Int = Math.ceil(Math.sqrt(Frames));
+		var cols:Int = Math.ceil(Frames / rows);
+		
+		var back = Clockwise ? FlxColor.WHITE : FlxColor.BLACK;
+		var fore = Clockwise ? FlxColor.BLACK : FlxColor.WHITE;
+		
+		var fullFrame = new FlxSprite().makeGraphic(W, H, FlxColor.TRANSPARENT, true);
+		var nextFrame = new FlxSprite().makeGraphic(W, H, back, false);
+		
+		if (Shape == SQUARE)
+		{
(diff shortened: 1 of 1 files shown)


flixel/addons/display/FlxPieDial.hx
```diff
@@ -0,0 +1,260 @@
...
+ * A dynamic shape that fills up the way a pie chart does. Useful for timers and other things.
+ * @author larsiusprime
+ */
+class FlxPieDial extends FlxSprite
+{
+
+	public var amount(default, set):Float = 0;
+	private var pieFrames:Int = 0;
+	
+	public static inline var CIRCLE:String = "circle";
```
Gama11: Any reason to use string constants over an enum for this?

Gama11: (the cache key isn't really a reason, enum values will convert to strings just fine when concatenated)

flixel/addons/display/FlxPieDial.hx
```diff
@@ -0,0 +1,260 @@
...
+	public static inline var SQUARE:String = "square";
+	
+	public function new(X:Float, Y:Float, Radius:Int, Color:FlxColor=FlxColor.WHITE, Frames:Int=36, Shape:String = CIRCLE, Clockwise:Bool = true, InnerRadius:Int=0)
```
Gama11: The Constructor has `X` and `Y`, but they're ignored - instead `0, 0` is passed to `super()`.

flixel/addons/display/FlxPieDial.hx
```diff
@@ -0,0 +1,260 @@
...
+import openfl.display.BlendMode;
+
+/**
+ * A dynamic shape that fills up the way a pie chart does. Useful for timers and other things.
+ * @author larsiusprime
+ */
+class FlxPieDial extends FlxSprite
+{
+
+	public var amount(default, set):Float = 0;
```
Gama11: What's the range for `amount`? Should probably be documented.

flixel/addons/display/FlxPieDial.hx
```diff
@@ -0,0 +1,260 @@
...
+		amount = f;
+		var frame:Int = Std.int(f * pieFrames);
+		animation.frameIndex = frame;
+		return amount;
+	}
+	
+	private function makePieDialGraphic(Radius:Int, Color:FlxColor, Frames:Int, Shape:String, Clockwise:Bool, InnerRadius:Int)
+	{
+		pieFrames = Frames;
+		var key:String = "pie_dial_" + Color.toHexString() + "_" + Std.string(Radius) + "_" + Frames + "_" + Shape;
```
Gama11: No need for `Std.string(Radius)` here within a string concatenation.

flixel/addons/display/FlxPieDial.hx
larsiusprime: 0.0-1.0

flixel/addons/display/FlxPieDial.hx
larsiusprime: I was using an enum originally, but it won't let you set an enum value as a default parameter, and I wanted it to default to circle. Know a way around this?

flixel/addons/display/FlxPieDial.hx
larsiusprime: whoops

flixel/addons/display/FlxPieDial.hx
adrianulima: May this work?
`?Shape:FlxPieDialShape`
```haxe
new(X:Float, Y:Float, Radius:Int, Color:FlxColor=FlxColor.WHITE, Frames:Int=36, ?Shape:FlxPieDialShape, ?Clockwise:Bool = true, ?InnerRadius:Int = 0)  {
_shape = (Shape != null) ? Shape : CIRCLE;
}
```
```haxe
enum FlxPieDialShape
- CIRCLE
- SQUARE
```

larsiusprime:
No crashes with latest hxcpp :)

Gama11:
If I comment out the `amount = 0;` from your demo code, the dial never shows up due to the missing setter call. That line shouldn't be necessary, the property is already initialized to 0.

flixel/addons/display/FlxPieDial.hx
```diff
@@ -0,0 +1,281 @@
...
+	public function new(X:Float, Y:Float, Radius:Int, Color:FlxColor=FlxColor.WHITE, Frames:Int=36, ?Shape:FlxPieDialShape, Clockwise:Bool = true, InnerRadius:Int=0) 
+	{
+		if (Shape == null) Shape = Circle;
+		super(X, Y);
+		makePieDialGraphic(Radius, Color, Frames, Shape, Clockwise, InnerRadius);
+	}
+	
+	public function set_amount(f:Float):Float
+	{
+		if (f < 0.0) f = 0.0;
```
Gama11: `FlxMath.bound()` would save a few lines of code here.

flixel/addons/display/FlxPieDial.hx
```diff
@@ -0,0 +1,281 @@
...
+			var bmp = makePieDialGraphicSub(Radius, Color, Frames, Shape, Clockwise, InnerRadius);
+			FlxG.bitmap.add(bmp, true, key);
+		}
+		
+		loadGraphic(key, true, W, H);
+		
+		amount = 1;
+	}
+	
+	private function makePieDialGraphicSub(Radius:Int, Color:Int, Frames:Int, Shape:FlxPieDialShape, Clockwise:Bool, InnerRadius):BitmapData
```
Gama11: This function is like 120 lines long which doesn't really help readability. Surely it could be split up a bit?

Gama11:
There is a subtle "flashing effect" in the animation right after the dial reaches 100% where it disappears for a frame. Very easy to see if you pause in the debugger and step ahead frame by frame.

larsiusprime:
@Gama11: pretty sure that is an artifact of tweening and has nothing to do with the actual underlying functionality.

flixel/addons/display/FlxPieDial.hx
larsiusprime: I just split it in two, better?

Gama11:
Ah, could be.

flixel/addons/display/FlxPieDial.hx
Gama11: They're still pretty huge, but yeah that's definitely better.

larsiusprime:
So how's it look?

larsiusprime:
@Gama11 anything else to do?

larsiusprime:
Fixed the last thing that travis was complaining about. Tested, should work now.

PR #198, opened 2015-10-07
Added support for imagelayers in Tiled loader, closes #155

### LINKED ISSUE #155: Add a "TiledImage.hx" to the Tiled library
Similar to this: 
https://github.com/kasoki/openfl-tiled-flixel/search?utf8=%E2%9C%93&q=ImageLayer&type=Code
https://github.com/kasoki/openfl-tiled/blob/fc82b84c7c49a5ec80a86cc5c62566542a8deb14/openfl/tiled/ImageLayer.hx

Right now (I think) the flixel-addons Tiled classes do not support Tiled's image layers, though such support can be very useful (for example, you can use Tiled as a non-tiles level editor).

sebbernery:
Fix issue #155 
Add support for imagelayer of Tiled ( here : http://doc.mapeditor.org/reference/tmx-map-format/ <imagelayer>)

### DIFF
--- a/flixel/addons/editors/tiled/TiledImageLayer.hx
+++ b/flixel/addons/editors/tiled/TiledImageLayer.hx
@@ -5,7 +5,7 @@ import haxe.xml.Fast;
 
 class TiledImageLayer extends TiledLayer
 {
-	public var image_path:String;
+	public var imagePath:String;
 	public var x:Int;
 	public var y:Int;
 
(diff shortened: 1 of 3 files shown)


Gama11:
Thanks!

Would be nice if this was used in the tiled demo somewhere, both to test it works correctly (now and in the future) and to show off the usage.

flixel/addons/editors/tiled/TiledImageLayer.hx
```diff
@@ -0,0 +1,20 @@
+package flixel.addons.editors.tiled;
+
+import haxe.xml.Fast;
+
+
+class TiledImageLayer extends TiledLayer
+{
+	public var image_path:String;
```
Gama11: Flixel uses camelCase, so this should be `imagePath`.

sebbernery:
Thank you for your feedbacks. I modified the Tiled demo to add an image layer : https://github.com/HaxeFlixel/flixel-demos/pull/204 .

sruloart:
Hooray! :)

PR #200, opened 2015-10-19
Transition refactor

larsiusprime:
Several improvements:

1. Fixed the animation timing synchronization for tiles. FlxTimer() wasn't producing reliable results, now I time things manually in update()
2. Allowed the user to specify framerate for tile transitions
3. Refactored so TileEffects can be used separately from transitions themselves (I need this functionality in Defender's Quest for the healing spell, for instance)

(I think I also accidentally pushed to a HaxeFlixel branch of "transition_refactor", feel free to delete that.)

### DIFF
--- a/flixel/addons/transition/TransitionTiles.hx
+++ b/flixel/addons/transition/TransitionTiles.hx
@@ -1,5 +1,7 @@
  package flixel.addons.transition;
+ 
 import flash.display.BitmapData;
+import flixel.addons.transition.TransitionEffect;
 import flixel.addons.transition.FlxTransitionSprite.TransitionStatus;
 import flixel.group.FlxGroup.FlxTypedGroup;
 
@@ -7,7 +9,7 @@ import flixel.group.FlxGroup.FlxTypedGroup;
  * 
  * @author larsiusprime
  */
-class TransitionTiles extends Transition
+class TransitionTiles extends TransitionEffect
 {
 	private var _grpSprites:FlxTypedGroup<FlxTransitionSprite>;
 	private var _isCenter:Bool = false;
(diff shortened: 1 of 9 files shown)


larsiusprime:
This is a pretty big refactor so I'm hesitant to merge it before a code review, whereas I think my PieDial changes are pretty simple by contrast. Would appreciate a code review anytime someone's ready.

@gamedevsam @gama11 @Beeblerox

flixel/addons/transition/TransitionTiles.hx
```diff
@@ -1,22 +1,25 @@
...
 
 /**
  * 
  * @author larsiusprime
  */
-class TransitionTiles extends Transition
+class TransitionTiles extends TransitionEffect
 {
-	private var _grpSprites:FlxTypedGroup<FlxTransitionSprite>;
+	private var _grpSprites:FlxTypedSpriteGroup<FlxTransitionSprite>;
```
Gama11: Why use a sprite group instead?

larsiusprime: Are you asking why TransitionEffect is a FlxSpriteGroup or why TransitionTiles _grpSprites member is?

With TransitionEffect it's a FlxSpriteGroup because TransitionEffects can now be used apart from state transitions; for instance in Defender's Quest I use a TransitionTiles object to drive a healing spell visual effect. It helps to have an object that behaves like a FlxSprite so that you an easily move the x/y position of the whole thing around.

As for _grpSprites itself I suppose on further reflection there's no reason it *needs* to be a FlxTypedSpriteGroup as its just an internal type. Do you have a preference in this case?

Gama11: I generally don't use sprite groups unless necessary - not much of a reason too, especially considering sprite groups are kind of a hack.

larsiusprime: Cool, if I revert that is everything else good?

larsiusprime:
@Gama11 Argh, now I remember. _grpSprites has to be a FlxSpriteGroup because you can't add() anything that's not FlxSprite-like to TransitionEffect, since that's now a FlxSpriteGroup too.

So I'm reverting the reversion. I do in fact need TransitionEffect to be a FlxSpriteGroup if it's going to be easily used outside of the state transition system.

PR #201, opened 2015-10-20
Fix pie dial

larsiusprime:
Fixes some lingering issues with FlxPieDial. Now it's accurate.

Demo is in this PR:
https://github.com/HaxeFlixel/flixel-demos/pull/206


larsiusprime:
I'm pretty sure this is good and ready for merging but should probably get a quick once over from a contributor. @gama11 or @gamedevsam got a second?

PR #204, opened 2015-10-23
FlxClothSprite new feature

adrianulima:
This class extends `FlxSprite` and draw it's frames in a mesh like a cloth. I think it can have a lot of use cases for creative users.

flxclothsprite_gif

It is already commented and working, but I think some parts of the code can be better.


flixel/addons/display/FlxClothSprite.hx
```diff
@@ -0,0 +1,533 @@
...
+					}
+				}
+				
+				_point.add(origin.x, origin.y);
+				if (isPixelPerfectRender(camera))
+				{
+					_point.floor();
+				}
+				
+				// Create a temporary frame and draw mesh's bitmapData
```
adrianulima: At this line I create a temporary frame and send it to camera. I do this because render mode is not FLX_RENDER_BLIT and the bitmapData sent is not used.

Have a better alternative to solve this?

Beeblerox: let me see what can be done with it

flixel/addons/display/FlxClothSprite.hx
```diff
@@ -0,0 +1,533 @@
...
+import flixel.util.FlxColor;
+import flixel.util.FlxDestroyUtil;
+import flixel.util.FlxSpriteUtil;
+import openfl.display.BitmapData;
+import openfl.display.Graphics;
+import openfl.geom.Point;
+import openfl.Vector;
+
+@:keep @:bitmap("assets/images/logo/default.png")
+private class GraphicDefault extends BitmapData {}
```
Gama11: Shouldn't include this again here, pretty sure that duplicates the graphic in memory. You can use `@:access` to access the original one.

flixel/addons/display/FlxClothSprite.hx
```diff
@@ -0,0 +1,533 @@
...
+ * A FlxSprite that draw it's frame in a mesh and behave like a cloth.
+ * 
+ * @author adrianulima
+ */
+class FlxClothSprite extends FlxSprite
+{
+	/**
+	 * An extra speed applied to mesh (in pixels per second). Use to simulate gravity or wind.
+	 */
+	public var gravity(default, null):FlxPoint = FlxPoint.get();
```
Gama11: `wind` might be a more fitting name considering we're dealing with a "cloth"? Also, `acceleration` is commonly used for gravity (in the phyiscs sense), so that could potentially be confusing.

`meshVelocity` might also work, following the pattern of `meshScale`?

flixel/addons/display/FlxClothSprite.hx
```diff
@@ -0,0 +1,533 @@
...
+	public var heightInTiles(default, null):Float;
+	
+	public var points(default, null):Array<ClothPoint> = [];
+	public var constraints(default, null):Array<ClothConstraint> = [];
+	
+	/**
+	 * Mesh arrays. Vertices, indices and uvtData to drawTriangles().
+	 */
+	#if flash
+	private var _vertices(default, null):Vector<Float>;
```
Gama11: You could use a typedef to avoid the redundancy here ([see here](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/graphics/tile/FlxDrawTrianglesItem.hx#L14)).

flixel/addons/display/FlxClothSprite.hx
```diff
@@ -0,0 +1,533 @@
...
+	/**
+	 * Change the size of the mesh related to the original frame. Need to call setMesh() to update.
+	 */
+	public var meshScale(default, null):FlxPoint = FlxPoint.get(1, 1);
+	/**
+	 * Bit field of flags (use with FlxObject.UP, DOWN, LEFT, RIGHT, NONE, ANY, etc) indicating pinned side. Use bitwise operators to check the values stored here.
+	 */
+	public var pinSide:Int;
+	/**
+	 * How many iterations will do on constraints for each update. Need to call setMesh() to update.
```
Gama11: The `Need to call setMesh() to update.` comment here seems wrong.

flixel/addons/display/FlxClothSprite.hx
```diff
@@ -0,0 +1,533 @@
...
+	 * 
+	 * @param	columns				Number of columns of the created mesh.
+	 * @param	rows				Number of rows of the created mesh.
+	 * @param   meshPixelsWidth		Optional, specify the width of the bitmapData where the mesh will be drawn.
+	 * 								It uses frameWidth by default, but when mesh state is bigger, a new BitmapData is created.
+	 * @param   meshPixelsHeight	Optional, specify the height of the bitmapData where the mesh will be drawn.
+	 * 								It uses frameHeight by default, but when mesh state is bigger, a new BitmapData is created.
+	 * @param   pinned				Indices of pinned points that are not affected by wind or velocity.
+	 */
+	public function setMesh(?columns:Int = 0, ?rows:Int = 0, ?meshPixelsWidth:Int = 0, ?meshPixelsHeight:Int = 0, ?pinned:Array<Int> = null):Void
```
Gama11: I don't think there's a reason for any of the `Int`s here to have a `?` ([making them `Null<Int>`](http://stackoverflow.com/questions/31914614/is-question-mark-operator-in-haxe-really-necessary/31916466#31916466)).

flixel/addons/display/FlxClothSprite.hx
adrianulima: I like the `meshVelocity` name, I was avoiding `acceleration` to not confuse with the original ones.
I think `gravity` seems more like vertical velocity and `wind` horizontal velocity. So I think `meshVelocity` is realy better.

flixel/addons/display/FlxClothSprite.hx
adrianulima: As you know I newbie with haxe, and never used a metadata pretty well. Can you show me how use `@:access` in this case.. xD

flixel/addons/display/FlxClothSprite.hx
adrianulima: I see, I will do that. But in this case that you show, it still need to instantiate it the same way I was doing? see line 39 of your link

flixel/addons/display/FlxClothSprite.hx
adrianulima: You're right, was a bad copy and paste.

flixel/addons/display/FlxClothSprite.hx
adrianulima: I copied it from other classes, I don`t know well what it means.. xD
If I take off all `?` it will works same way?

flixel/addons/display/FlxClothSprite.hx
adrianulima: Thanks @Beeblerox, if we can make this better I was thinking in a better way to do effects like `FlxWaveSprite`, `FlxGlitchSprite`, and others effects.

I've started the code, but I will back coding only on monday. My idea is to create a new class called `FlxEffectSprite` that have an array called `effects`, so we will add `FlxWaveEffect` and `FlxGlitchEffect` that will modify the `bitmapData` just before draw it like I do in this class.
This `Flx___Effect` classes will need just a bitmap modifier method. This way effects will work like bitmap filters and will be easily chainned.
I do not know if I explained well, I can show you next week.
What you guys think?

Beeblerox: @adrianulima i think i've understood that you're talking about. But i thought it would be better to extend `FlxStrip` class which already has some drawTriangles related functionality. Not ready to tell what exactly is better.
But anyaway, you're doing great work!

Beeblerox: and the idea with chainable effects seems good to me

adrianulima: Wow this `FlxStrip` has some functionalities that I made in this `FlxClothSprite` class. Maybe I should extend `FlxStrip` and use it's `draw` method if it works better than I'm using.
This is what you said?

Beeblerox: Yes, it is. But `FlxStrip` doesn't use `frameBitmap` for rendering. Maybe we could change it to use it, so it will make chains of effect possible

flixel/addons/display/FlxClothSprite.hx
```diff
@@ -0,0 +1,517 @@
...
+				_point.add(_drawOffset.x, _drawOffset.y);
+				if (isPixelPerfectRender(camera))
+				{
+					_point.floor();
+				}
+				
+				// Create a temporary frame and draw mesh's bitmapData
+				var frameGraphic:FlxGraphic = FlxGraphic.fromBitmapData(framePixels, true, null, false);
+				
+				camera.drawTriangles(frameGraphic, _vertices, _indices, _uvtData, colors, _point, blend, antialiasing);
```
adrianulima: @Beeblerox 
I'm using `camera.drawTriangles()` here the same way is used on `FlxStrip`. I didn't extended it because I prefer keep _vertices, _incides and _uvtData private, and I would rewrite almost everything on `draw()` method.
For flash `camera.drawTriangles()` is slow. So I still using `camera.drawPixels()`.

gamedevsam:
This is something I'd like to see for 4.0 release, what updates need be made before we merge this?

Gama11:
How does this relate to #205, shouldn't it be an `IFlxEffect` as well?

flixel/addons/display/FlxClothSprite.hx
```diff
@@ -0,0 +1,518 @@
...
+			updateConstraints(elapsed);
+		}
+		
+		super.update(elapsed);
+	}
+	
+	/**
+	 * Called by game loop, updates then blits or renders current frame of animation to the screen.
+	 */
+	override public function draw():Void
```
Gama11: Please check if it's sufficient to override `drawComplex()`. This creates a lot of redundant code.

adrianulima: Wow, good work with https://github.com/HaxeFlixel/flixel/commit/6d3d349c3c5f942f741e2522f7ddf1e300742193

I will look at it and try to eliminate some redundance.

adrianulima:
This could be part of https://github.com/HaxeFlixel/flixel-addons/pull/205 but I think it work better separeted because of `update()` method, I don't remember other reasons now. I will take a look at it and see if it can be an `IFlxEffect`

adrianulima:
I did the update of `draw()` override. Also kept not a `IFlxEffect`.

The demo still the same, and should be working. https://github.com/HaxeFlixel/flixel-demos/pull/207

Gama11:
Nice, that looks much cleaner.

flixel/addons/display/FlxClothSprite.hx
```diff
@@ -0,0 +1,484 @@
...
+			FlxSpriteUtil.flashGfx.beginBitmapFill(framePixels, null, false, true);
+			FlxSpriteUtil.flashGfx.drawTriangles(_vertices, _indices, _uvtData);
+			FlxSpriteUtil.flashGfx.endFill();
+			
+			this.meshPixels.draw(FlxSpriteUtil.flashGfxSprite);
+		}
+	}
+}
+
+typedef ClothPoint = {
```
Gama11: These typedefs are missing the `Flx`-prefix.

adrianulima:
Added the `Flx` prefix. And is there a problem to use `_frameGraphic` as I'm using it? Or a better way to use it?

Gama11:
No idea, seems to be working well enough though.

I like the custom drawDebug!

Gama11:
Allocations could probably be optimized a bit, I don't think they shouldn't happen this frequently:

PR #205, opened 2015-10-26
FlxEffectSprite effects refactor

adrianulima:
Refact how effects are applied to sprites, now it can easily chain effects and avoid a lot of `FlxSprite` instances.
```haxe
var target:FlxSprite = new FlxSprite();
target.makeGraphic(250, 250);

var obj:FlxEffectSprite = new FlxEffectSprite(target);
add(obj);

var eff1:FlxRainbowEffect = new FlxRainbowEffect(0.5);		
var eff2:FlxWaveEffect = new FlxWaveEffect(FlxWaveMode.ALL, 40, -1, 5, 20);
var eff3:FlxGlitchEffect = new FlxGlitchEffect(10, 2, 0.1);

obj.effects.push(eff1);		
obj.effects.push(eff2);
obj.effects.push(eff3);
```
capeffectalpha

```haxe
FlxRainbowEffect, FlxOutlineEffect, FlxWaveEffect, FlxGlitchEffect, FlxShakeEffect, FlxTrailEffect
```
capeffects_160


flixel/addons/effects/FlxEffectSprite.hx
```diff
@@ -0,0 +1,174 @@
...
+				{
+					_point.floor();
+				}
+				
+				// Create a temporary frame and draw effect's bitmapData
+				var frameEffect:FlxFrame = new FlxFrame(FlxGraphic.fromBitmapData(_effectPixels, true, null, false));
+				frameEffect.frame = new FlxRect(0, 0, _effectPixels.width, _effectPixels.height);
+				
+				_matrix.translate(_point.x + _drawOffset.x, _point.y + _drawOffset.y);
+				camera.drawPixels(frameEffect, framePixels, _matrix, cr, cg, cb, alpha, blend, antialiasing);
```
adrianulima: @Beeblerox This is the same trick I was using in FlxClothSprite, but there I could change for `camera.drawTriangles()`. Here I don't know a better way to do this.

adrianulima:
I added `FlxEffectRainbow` but to work I added it first. I think `bitmapData.copyPixels()` alpha mask is not working (only for flash). The original `FlxRainbowSprite` has the same problem.

adrianulima:
I added `FlxEffectOutline`, but it is too heavy to be updated every frame, so it draws just once unless you set `dirty` true again.
This way it should be added to the sprite before effects that change the shape of the `bitmapData` like wave and glitch.

Gama11:
Wow, this is definitely a lot more elegant of a solution than #173! :+1: 

I guess this depends on #204 being merged first (there are merge conflicts)?

adrianulima:
@Gama11 there are no conflicts..
I could create an `FlxEffectCloth` to reproduce cloth and everything, but FlxClothSprite uses `camera.drawTriangles()` and some physics of `FlxSprite`.
So they are separated features.

Tomorrow I will commit three more effects, `FlxEffectFade`, `FlxEffectFlash` and `FlxEffectShake`. They will work similar to `fade`, `flash` and `shake` of `FlxCamera`.

Maybe in the future we could also apply this `FlxEffects` to `FlxCamera` or create a `FlxEffectCamera`, so would be possible apply `rainbow`, `glitch`, etc to cameras.

Gama11:
>there are no conflicts..

adrianulima:
xD I don`t know what it is, maybe is because I deleted some classes?

adrianulima:
I added `FlxEffectFade` based on `FlxCamera` fade. I didn't understand why `FlxCamera` also have `flash()` and `fade()` method, both seems to do the same effect. Am I wrong? https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxCamera.hx#L899

@Gama11 about the conflict my guess is because I deleted `FlxGlitchSprite` and `FlxWaveSprite`. I could have done this? Should I commit it back?

adrianulima:
When I deleted `FlxGlitchSprite` and `FlxWaveSprite`they were outdated, so commit conflicted. I updated my branch to Haxeflixel/dev and seems the conflicts gone.

Should I delete these classes again?

adrianulima:
Added `FlxEffectShake` and `FlxEffectTrail`. When(If) this branch be merged we'll can create more effects, for now I will stop committing new ones.
I will update PR description with a new gif and the list of effects.

flixel/addons/effects/FlxEffectSprite.hx
```diff
@@ -0,0 +1,221 @@
...
+	 * Core update loop
+	 */
+	override public function update(elapsed:Float):Void
+	{
+		if (updateEffects)
+		{
+			#if !FLX_RENDER_BLIT
+			getFlxFrameBitmapData();
+			#end
+			effectPixels = framePixels.clone();
```
Beeblerox: i think we need to dispose effectPixels that already exist before this line

flixel/addons/effects/FlxEffectSprite.hx
```diff
@@ -0,0 +1,221 @@
...
+				
+				_point.add(origin.x, origin.y);
+				if (isPixelPerfectRender(camera))
+				{
+					_point.floor();
+				}
+				
+				// Create a temporary frame and draw effect's bitmapData
+				var frameEffect:FlxFrame = new FlxFrame(FlxGraphic.fromBitmapData(effectPixels, true, null, false));
+				frameEffect.frame = new FlxRect(0, 0, effectPixels.width, effectPixels.height);
```
Beeblerox: these two lines could possibly create a lot of garbage. Let me think how they could be changed

flixel/addons/effects/FlxEffectSprite.hx
```diff
@@ -0,0 +1,221 @@
...
+				
+				_matrix.translate(_point.x + _effectOffset.x, _point.y + _effectOffset.y);
+				camera.drawPixels(frameEffect, framePixels, _matrix, cr, cg, cb, alpha, blend, antialiasing);
```
Beeblerox: i think that you meant not `framePixels` here but `effectPixels` (like in simple render case with copyPixels operation)

adrianulima: You're right, but the `camera.drawPixels()` don't use the `bitmapData` passed, or at least wasn't using.
But it's better pass `effectPixels` anyway, next commit I change this.

flixel/addons/effects/FlxEffectSprite.hx
adrianulima: @Beeblerox do you found a better way to do this part? I'm very busy lately, but I'm still here if you need anything

gamedevsam:
Any updates on this change? This seems like something we'd want for 4.0.

adrianulima:
**I'm here** if you guys need me to change or improve those classes. It was made months ago and a lot of code has changed, may need to test it again.
@Beeblerox can know better what we need here.

flixel/addons/effects/IFlxEffect.hx
```diff
@@ -0,0 +1,12 @@
+package flixel.addons.effects;
+import openfl.display.BitmapData;
+import openfl.geom.Point;
+
+interface IFlxEffect
+{
+	public var active:Bool;
+	public var offset:Point;
+	public function destroy():Void;
```
Gama11: This line should be removed, instead this interface should extend `IFlxDestroyable`.

flixel/addons/effects/FlxRainbowSprite.hx
```diff
@@ -1,90 +0,0 @@
...
-import flixel.util.FlxColor;
-import openfl.display.BitmapData;
-
-/**
- * This is a FlxSprite that copies a target FlxSprite and creates a color-cycling effect, using the target's graphic as a mask.
- * Usage: Create a new FlxSprite, as normal, and position it where you want. Create a new FlxRainbowSprite and pass it your target,
- * and add it to the state. 
- * @author Tim Hely / tims-world.com
- */
-class FlxRainbowSprite extends FlxSprite
```
Gama11: Is there a reason you only removed `FlxRainbowSprite`, but not `FlxGlitchSprite`, `FlxWaveSprite` and `FlxOutline`, even though there are `IFlxEffect` implementations of them?

flixel/addons/effects/FlxEffectTrail.hx
```diff
@@ -0,0 +1,188 @@
...
+import openfl.geom.Matrix;
+import openfl.geom.Point;
+using flixel.util.FlxArrayUtil;
+
+/**
+ * This creates a trail copying the bitmapData many times.
+ * 
+ * @author adrianulima
+ */
+class FlxEffectTrail implements IFlxEffect
```
Gama11: What advantage does this have over the already existing `FlxTrail` and `FlxTrailArea`? Those are fairly flexible already, and having 3 separate implementations for essentially the same thing seems like overkill.

flixel/addons/effects/FlxEffectSprite.hx
Gama11: Doesn't look like this has been changed yet?

flixel/addons/effects/FlxEffectFade.hx
```diff
@@ -0,0 +1,154 @@
...
+import flixel.util.FlxDestroyUtil;
+import openfl.display.BitmapData;
+import openfl.geom.Point;
+
+/**
+ * This gradually fills the bitmapData with a color.
+ * 
+ * @author adrianulima
+ */
+class FlxEffectFade implements IFlxEffect
```
Gama11: I'm not sure this is really all that useful, can't you achieve pretty much the same effect by tweening the `.color` property?

Gama11:
Regarding the class names, I think it would be more natural to name them `FlxRainbowEffect`, `FlxWaveEffect` etc instead of the other way around.

flixel/addons/effects/FlxEffectSprite.hx
adrianulima: Not yet, I was waiting next comments before work on it again. I think some things changed on `camera.drawPixels()`, so I have to look at it again.

flixel/addons/effects/FlxRainbowSprite.hx
adrianulima: https://github.com/HaxeFlixel/flixel-addons/pull/205#issuecomment-151832324
I removed them, but it caused conflicts so I commited it back

flixel/addons/effects/FlxEffectTrail.hx
adrianulima: `FlxTrail` use an array of `FlxSprite` and `FlxEffectTrail` use an array of `FlxPoint`. I thought it might be lighter with points.

`FlxEffectTrail` can be in the middle of `FlxEffectSprite#effects` chain, this may have some use cases.

flixel/addons/effects/FlxEffectFade.hx
adrianulima: I basically used the same approach as `FlxCamera#fade`.

and also `FlxEffectFade` can be in the middle of `FlxEffectSprite#effects` chain, this may have some use cases.

adrianulima:
I can rename them.
I did like that to group them when imported and when sorting by name.

flixel/addons/effects/IFlxEffect.hx
adrianulima: Ok, I will make all this changes today, and review everything to see what has changed and what can be improved here.

flixel/addons/effects/FlxEffectTrail.hx
Gama11: `FlxTrailArea` should be pretty light, it renders everything to a single Bitmap. Maybe it should be removed if they're similar enough..

Gama11:
The problem is that it's inconsistent with the naming pattern we currently use - we have `FlxNapeSprite`, `FlxRainbowSprite` etc - not `FlxSpriteNape`, `FlxSpriteRainbow`.

flixel/addons/effects/FlxEffectTrail.hx
adrianulima: Ok. I will take a look in `FlxTrailArea` too, and choose the best approach to apply on `FlxEffectTrail`

Gama11:
Any update here?

adrianulima:
Nothing yet, was too busy the last few weeks. But I have already installed Windows, FD and Haxe on my personal computer and I have 4 days off from work and I'll just fix this classes.

flixel/addons/effects/FlxEffectFade.hx
adrianulima: @Gama11 Do you think it worth send `FlxFadeEffect` together in this pull? Since tween aparently works easier.

flixel/addons/effects/FlxRainbowSprite.hx
adrianulima: @Gama11 should I delete again `FlxFlitchSprite` and `FlxWaveSprite`? Also `FlxOutline`?

adrianulima:
All planned changes were made, when this PR is ready and closed I will create a demo for it.

flixel/addons/effects/FlxRainbowSprite.hx
Gama11: Yes, we don't want to maintain two versions of everything.

flixel/addons/effects/FlxEffectSprite.hx
```diff
@@ -0,0 +1,121 @@
...
+	
+	override public function getScreenPosition(?point:FlxPoint, ?Camera:FlxCamera):FlxPoint 
+	{
+		return super.getScreenPosition(point, Camera).addPoint(_effectOffset);
+	}
+	
+	/**
+	 * Core update loop
+	 */
+	override public function update(elapsed:Float):Void
```
Gama11: This should probably be done in a `draw()` override, this is all rendering-related.

flixel/addons/effects/FlxEffectSprite.hx
```diff
@@ -0,0 +1,121 @@
...
+	{
+		if (effectsEnabled)
+		{
+			#if !FLX_RENDER_BLIT
```
Gama11: The `FLX_RENDER_BLIT` define doesn't exist anymore (see https://github.com/HaxeFlixel/flixel/pull/1668).

flixel/addons/effects/FlxTrailEffect.hx
```diff
@@ -0,0 +1,242 @@
...
+	
+	private function disposeCachedPixels() 
+	{
+		while(_recentPixels.length > 0) 
+		{
+			FlxDestroyUtil.dispose(_recentPixels.shift());
+		}
+	}
+	
+	public function set_length(Value:Int):Int
```
Gama11: setters should always be private.

flixel/addons/effects/FlxTrailEffect.hx
```diff
@@ -0,0 +1,242 @@
...
+			if (_recentPixels.length > _recentPositions.length)
+			{
+				_recentPixels.shift();
+			}
+		}
+		
+		return length = Value;
+	}
+	
+	public function set_cachePixels(Value:Bool):Bool
```
Gama11: setters should always be private.

flixel/addons/effects/IFlxEffect.hx
```diff
@@ -0,0 +1,12 @@
+package flixel.addons.effects;
+import flixel.util.FlxDestroyUtil.IFlxDestroyable;
+import openfl.display.BitmapData;
+import openfl.geom.Point;
+
+interface IFlxEffect extends IFlxDestroyable
+{
+	public var active:Bool;
+	public var offset:Point;
```
Gama11: This should be a `FlxPoint`.

flixel/addons/effects/FlxEffectSprite.hx
```diff
@@ -0,0 +1,121 @@
...
+	
+	/**
+	 * Creates a FlxEffectSprite at a specified position with a specified one-frame graphic. 
+	 * If none is provided, a 16x16 image of the HaxeFlixel logo is used.
+	 * 
+	 * @param	X				The initial X position of the sprite.
+	 * @param	Y				The initial Y position of the sprite.
+	 * @param	SimpleGraphic	The graphic you want to display (OPTIONAL - for simple stuff only, do NOT use for animated images!).
+	 */
+	public function new(Target:FlxSprite, ?Effects:Array<IFlxEffect> = null)
```
Gama11: The `= null` is redundant if you already have the `?`, see [here](http://stackoverflow.com/questions/31914614/is-question-mark-operator-in-haxe-really-necessary/31916466#31916466).

flixel/addons/effects/FlxEffectSprite.hx
```diff
@@ -0,0 +1,121 @@
...
+		_effectOffset = FlxPoint.get();
+	}
+	
+	/**
+	 * WARNING: This will remove this sprite entirely. Use kill() if you want to disable it temporarily only and reset() it later to revive it.
+	 * Used to clean up memory.
+	 */
+	override public function destroy():Void
+	{
+		for (effect in effects)
```
Gama11: Line 66-71 can be replaced with `effects = FlxDestroyUtil.destroyArray(effects);`

flixel/addons/effects/FlxShakeEffect.hx
```diff
@@ -0,0 +1,129 @@
...
+	
+	public function destroy():Void 
+	{
+		offset = null;
+		onComplete = null;
+	}
+	
+	public function update(elapsed:Float):Void 
+	{
+		if (_time > 0)
```
Gama11: You could invert this if to reduce the nesting of the function:

```haxe
if (_time == 0)
    return;
```

flixel/addons/effects/FlxTrailEffect.hx
```diff
@@ -0,0 +1,242 @@
...
+				_recentPixels.push(target.pixels.clone());
+			}
+			
+			_currentFrames = 0;
+		}
+	}
+	
+	public function apply(bitmapData:BitmapData):BitmapData 
+	{
+		if (_recentPositions.length >= 1)
```
Gama11: This is a good candidate for inversion as well.

flixel/addons/effects/FlxTrailEffect.hx
```diff
@@ -0,0 +1,242 @@
...
+			
+			return _pixels.clone();
+		}
+		
+		return bitmapData;
+	}
+	
+	private function disposeCachedPixels() 
+	{
+		while(_recentPixels.length > 0)
```
Gama11: Missing space after `while`.

adrianulima:
Thanks for comments, I fixed all that. Only doubt was about move rendering code to `draw`, this cause a need to iterate `effects` twice, at `update` and `draw` methods. If this is not a problem, it is done.
I also deleted `FlxGlitchSprite`, `FlxWaveSprite` and `FlxOutline`.

But didn't deleted `FlxTrail` and `FlxTrailArea` because `FlxTrailEffect` works different from them.
I don't know if `FlxTrail` and `FlxTrailArea` are working properly, might be good someone that know this classes test them.

Gama11:
Iterating over a small array is pretty neglegible in terms of performance, especially when compared to say, rendering.

>I don't know if FlxTrail and FlxTrailArea are working properly

Why, is there anything that makes you think they aren't working properly?

Gama11:
It might make sense to put these new classes into a `flixel.addons.effects.chainable` subpackage (there might be better names than `chainable`), just so they're grouped together and it's clear they belong together. Otherwise, it might be quite confusing why there's a `flixel.addons.effects.FlxTrail`, but also a `flixel.addons.effects.FlxTrailEffect`.

What do you think?

adrianulima:
I tried to use `FlxTrail` with an animated `FlxSprite` as I do with `FlxTrailEffect`. But it changes one frame per `draw`.
I don't know if I'm doing something wrong:
```haxe
tgt = new FlxSprite();
tgt.loadGraphic("assets/images/flags.png", true, 140, 140);
tgt.animation.add("anim", [0, 1, 2, 3], 3, true);
tgt.animation.play("anim");
		
obj = new FlxTrail(tgt,null,10,1);
add(obj);
add(tgt);//Do I have to add both?

override public function update(elapsed:Float):Void
{
	super.update(elapsed);
	tgt.x = FlxG.mouse.x;
	tgt.y = FlxG.mouse.y;
}
```
trail

Gama11:
Hm.. yeah, that looks broken. I've never really tested FlxTrail with animations, I believe support for it was added at some point. Maybe some API change has broken it since then.

flixel/addons/effects/FlxEffectSprite.hx
```diff
@@ -0,0 +1,134 @@
...
+import openfl.geom.Point;
+
+/**
+ * Use this class to chain bitmap effects to your target FlxSprite frame.
+ * @author adrianulima
+ */
+class FlxEffectSprite extends FlxSprite
+{
+	/**
+	 * Enables and disables the effects update loop.
```
Gama11: This comment is a bit misleading now, it's not just the update loop, also draw.

Gama11: Besides, is it really necessary to have this flag? There are already active and visible to control whether update and raw are run respectively.

Gama11:
Could you update the demos that are failing on Travis with these changes?

flixel/addons/effects/chainable/FlxEffectSprite.hx
```diff
@@ -0,0 +1,134 @@
...
+	
+	override public function draw():Void 
+	{
+		target.drawFrame();
+		
+		if (target.framePixels == null)
+			return;
+		
+		FlxDestroyUtil.dispose(pixels);
+		pixels = target.framePixels.clone();
```
Gama11: This seems quite expensive, does it really need to happen every draw() call?

adrianulima: I will test other solutions here...

Gama11:
Thanks! :+1:

Issue #210, opened 2015-12-01
Tiled: Support for tilesets made with collection of images


Tembac:
I want to be able to do something like this video using HaxeFlixel and Tiled. 

https://www.youtube.com/watch?v=yVgcFdHOl6w

I think that what needs to be improved is the support of tiles that are made of collection of images. If I try to use that kind of tilesets now, HaxeFlixel Crashes. 

I'm trying to make the necessary changes to support this. 

@gamedevsam, you made the initial implementation of tiled. Do you have any advised on how to approach this new feature?

PR #212, opened 2015-12-03
Tiled: Support for tilesets made with collection of images

Tembac:
I made the changes proposed here: https://github.com/HaxeFlixel/flixel-addons/issues/210

It works fine. It will be great if tiled supports depth sorting as I'm asking here: https://github.com/bjorn/tiled/issues/888

I'm still getting used to github. I hope I don't mess anything.


flixel/addons/editors/tiled/TiledImageSet.hx
```diff
@@ -0,0 +1,51 @@
+package flixel.addons.editors.tiled;
+
+import haxe.xml.Fast;
+
+/**
+ * Copyright (c) 2013 by Samuel Batista
+ * (original by Matt Tuttle based on Thomas Jahn's. Haxe port by Adrien Fischer)
+ * This content is released under the MIT License.
+ */
+class TiledImageSet implements Dynamic<String>
```
Gama11: This is a 1:1 copy of `TiledPropertySet` except for the implementation of `extend()` - a common base class might not be a bad idea.

flixel/addons/editors/tiled/TiledTileSet.hx
```diff
@@ -59,61 +61,139 @@ class TiledTileSet
...
+			if (!source.hasNode.image)
 			{
-				if (!node.has.id)
+				//tileset with several images
+				var node:Fast = source.node.tile;
+				imageSource = "";
+				
+				name = source.att.name;
+				
+				if (source.has.tilewidth)
```
Gama11: Again, there seems to be a fair amount of redundancy here (this section is quite similar to the one for "normal tileset").

flixel/addons/editors/tiled/TiledTileSet.hx
```diff
@@ -59,61 +61,139 @@ class TiledTileSet
...
+					tileProps[id].keys.set("id", Std.string(id));
+					for (prop in node.nodes.properties)
+					{
+						tileProps[id].extend(prop);
+					}
+				}
+				
+				if (tileWidth > 0 && tileHeight > 0)
+				{
+					numRows = cast(imgWidth / tileWidth);
```
Gama11: `cast` is a bad idea for float-to-int conversion, `Std.int()` is safer.

flixel/addons/editors/tiled/TiledTileSet.hx
```diff
@@ -25,9 +25,11 @@ class TiledTileSet
 	public var numCols:Int;
 	
 	public var properties:TiledPropertySet;
-	
 	public var tileProps:Array<TiledPropertySet>;
 	
+	public var imagesSources:TiledImageSet;
```
Gama11: Should be `imageSources` and `tileImageSources`, the duplicate plural looks strange.

flixel/addons/editors/tiled/TiledTileSet.hx
```diff
@@ -59,61 +61,139 @@ class TiledTileSet
...
+					
+					var id:Int = Std.parseInt(node.att.id);
+					tileImagesSources[id] = new TiledImageSet();
+					tileImagesSources[id].keys.set("id", Std.string(id));
+					tileImagesSources[id].extend(node);
 				}
+				
+				if (tileWidth > 0 && tileHeight > 0)
+				{
+					numRows = 0;//cast(imgWidth / tileWidth);
```
Gama11: What's with the commented out code here?

Gama11:
The tiled demo should have an example of this feature to be able to verify it works correctly.

Tembac:
Thanks for the feedback @Gama11. I'll work on it. 

Also, Tiled has a way to sort objects by depth that I didn't knew so I'll see if I can implement that as well.

flixel/addons/editors/tiled/TiledImageSet.hx
Tembac: I was not really sure what "Dynamic<String>" is so I made another class just in case. It is better if I use the same class and add another function to add the image properties?

Gama11: Not sure, what does a `TiledImageSet` look like in the xml compared to a `TiledImageSet`?

Tembac: This is a property:
 <tile id="0">
   <properties>
    <property name="type" value="player"/>
   </properties>
  </tile>

This is an image:
 <tile id="0">
   <image width="769" height="720" source="../images/backgrounds/pack2/capa01_A.png"/>
  </tile>

Properties only have two parameters and image has three.

Gama11: That doesn't really look like a "set" to me. What makes That a `TiledPropertySet` is that there are multiple `<property>` nodes in `<properties>`. `<image>` should probably just be `TiledImage` with `width`, `height` and `source`.

Gama11: Or is there a way to have a tile with multiple images?

Gama11:
Closing since this has merge conflicts by now and won't be merged in its current state. Feel free to open a new PR with an improved version.

Tembac:
:+1: 
Yes, I want to finish this as soon as possible but I was finishing a job and now I'm sick :(
I should make a demo also.

Gama11:
Instead of creating an entirely new demo, it probably makes sense to expand / improve the existing Tiled Editor demo.

PR #213, opened 2015-12-05
Minor cleanup that helps with memory leaks

larsiusprime:


larsiusprime:
Gonna go ahead and merge this as it's super tiny and seems safe.

PR #216, opened 2015-12-11
Fix travis

larsiusprime:
These changes fix the three demos that are failing on Travis, holding up my runtime render refactor.


larsiusprime:
Since:

- These are such minor changes
- I've tested them locally and they fix the previously broken builds
- Travis is green

I'm gonna go ahead and merge them to save us all some time. That okay?

PR #217, opened 2015-12-17
fixing up screengrab dialogs to have no external dependencies

larsiusprime:
(just uses lime/openfl now)

Gonna wait on travis and see how this goes

### DIFF
--- a/flixel/addons/plugin/screengrab/FlxScreenGrab.hx
+++ b/flixel/addons/plugin/screengrab/FlxScreenGrab.hx
@@ -1,7 +1,9 @@
 package flixel.addons.plugin.screengrab;
 
-#if (sys && systools)
-import systools.Dialogs;
+#if sys
+import lime.ui.FileDialog;
+import lime.ui.FileDialogType;
+import openfl.display.PNGEncoderOptions;
 #end
 
 #if !js
@@ -14,7 +16,7 @@ import flixel.addons.util.PNGEncoder;
 import flixel.FlxG;
 import flixel.input.keyboard.FlxKey;
 
-#if flash
+#if !sys
 import flash.net.FileReference;
 #end
 
@@ -136,13 +138,8 @@ class FlxScreenGrab extends FlxBasic
 		return theBitmap;
 	}
 	
-	private static function save(Filename:String = ""):Void
+	private static function fixFilename(Filename:String):String
 	{
-		if (screenshot.bitmapData == null)
-		{
-			return;
-		}
-		
 		if (Filename == "")
 		{
 			var date:String = Date.now().toString();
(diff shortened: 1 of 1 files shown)


larsiusprime:
NOTE: bitmapdata.draw() on the stage doesn't currently work on native targets. I have an open PR to openfl/lime to add native screenshots, once that's pulled in I can update this so not only are the dialogs prettier, but screenshots will actually work on native targets.

flixel/addons/plugin/screengrab/FlxScreenGrab.hx
```diff
@@ -155,28 +152,54 @@ class FlxScreenGrab extends FlxBasic
...
+		{
+			return;
+		}
+		
+		Filename = fixFilename(Filename);
+		
+		var png:ByteArray;
+	#if flash
+		png = PNGEncoder.encode(screenshot.bitmapData);
+	#elseif openfl_legacy
```
Gama11: Is `openfl_legacy` still being defined?

Gama11:
Great, one less dependency! :+1:

MSGhero:
Good. There was missing parity between systools' cpp saving and flash's, and the lib had never been updated to correct that (there is an optional parameter in the cpp that systools wouldn't let you use).

flixel/addons/plugin/screengrab/FlxScreenGrab.hx
larsiusprime: Yes. There is a openfl_legacy and a lime_legacy define.

Gama11: Is there a difference between them? I think we use lime_legacy everywhere else, so that might make sense for consistency.

larsiusprime: There is a difference. IIRC, Joshua told me that if you're trying to test against legacy/next, then you should use "openfl_legacy" if the thing you're guarding against is in an openfl. or flash. package, and "lime_legacy" if it's in an lime. package.

larsiusprime: Oh I think I remember now. It's possible, with lime "hybrid" mode, to have the new lime backend but the old openfl, or vice versa, I forget. But it's possible for them to be mismatched so it is smart to articulate the one you actually mean.

Gama11: Gotcha. We're probably "doing it wrong" in a few other places then.

Gama11:
Hm, I guess the FileBrowse demo still has systools as a dependency, so I can't really remove it from travis yet...

PR #219, opened 2016-01-20
Fix some memory leaks (FlxTransitionableState was a HUGE one)

larsiusprime:


flixel/addons/display/shapes/FlxShapeArrow.hx
```diff
@@ -77,11 +77,15 @@ class FlxShapeArrow extends FlxShape
...
 	}
 	
-	override public function destroy():Void
+	override public function destroy():Void 
 	{
+		FlxDestroyUtil.destroy(point);
+		FlxDestroyUtil.destroy(point2);
+		FlxDestroyUtil.destroyArray(_vertices);
+		_vertices = null;
+		_matrix2 = _matrix;
```
gamedevsam: I don't understand this here, why assign matrix to eachother in destroy?

larsiusprime: supposed to be null.

larsiusprime:
Fixed, got my wires crossed.

Gama11:
So the destroy functions actually have a measurable impact? There was always some debate whether they're still necessary with modern GC.

larsiusprime:
I don't know if the destroy functions have any impact or not... honestly the main issues in this function are the transition bits, skipping the super and leaving the callback intact where causing entire states to persist.

Gama11:
I see, makes sense.

MSGhero:
Lol, I meant the syntax to destroy is 'thing = FlxDestroyUtil.destroy(thing)', although I usually just destroy things myself and set them to null after.

PR #220, opened 2016-01-21
Changed Int on UInt.

n06rin:
If we try to build on flash, we have problem:
    UInt might lead to unexpected results
But if we use (mapData.position : UInt) everything is ok.

### DIFF
--- a/flixel/addons/editors/tiled/TiledTileLayer.hx
+++ b/flixel/addons/editors/tiled/TiledTileLayer.hx
@@ -181,8 +181,8 @@ class TiledTileLayer extends TiledLayer
 			}
 
 			tileArray = new Array<Int>();
-
-			while ((mapData.position : Int) < mapData.length)
+			
+			while ((mapData.position : UInt) < mapData.length)
 			{
 				tileArray.push(resolveTile(mapData.readUnsignedInt()));
 			}


Gama11:
What are your Flixel / OpenFL / Lime / Haxe versions?

n06rin:
flixel-addons: 1.1.0 git [dev:C:\Development\HaxeToolkit\haxe\lib\flixel-addons/git]
flixel: 3.3.11 git [dev:C:\Development\HaxeToolkit\haxe\lib\flixel/git]
haxelib_client: [3.1.0-rc.4] 3.2.0-rc.3
lime: [2.7.0] 2.8.1
openfl-samples: [3.3.1]
openfl: [3.4.0] 3.5.1 3.5.3
swf: [2.1.3]

Gama11:
This issue is specific to older OpenFL (pre 3.5) versions before the `ByteArray` refactor. Your fix only flips the issue around and causes a compiler error with new OpenFL versions (see Travis).

n06rin:
Oh, okay.
Can you say, latest stable openfl version is okay for dev  version of haxeflixel?

Gama11:
Sure. As you just noticed, it's probably more stable because we test it on Travis, unlike older OpenFL versions.

I just pushed a fix for this that should be compatible with all OpenFL versions though.

PR #223, opened 2016-01-26
Class variables more accessible

AustinEast:
Made it easier to edit both the speed and margin of the zoom on creation of the object, and added functions to change them in real time.

### DIFF
--- a/flixel/addons/display/FlxZoomCamera.hx
+++ b/flixel/addons/display/FlxZoomCamera.hx
@@ -29,7 +29,7 @@ class FlxZoomCamera extends FlxCamera
 	 */
 	private var _zoomMargin:Float;
 	
-	public function new(X:Int, Y:Int, Width:Int, Height:Int, Zoom:Float = 0)
+	public function new(X:Int, Y:Int, Width:Int, Height:Int, Zoom:Float = 0, ZoomSpeed:Float = 25, ZoomMargin:Float = 0.25)
 	{
 		super(X, Y, Width, Height, FlxCamera.defaultZoom);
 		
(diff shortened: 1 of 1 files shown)


flixel/addons/display/FlxZoomCamera.hx
```diff
@@ -29,7 +29,7 @@ class FlxZoomCamera extends FlxCamera
 	 */
 	private var _zoomMargin:Float;
 	
-	public function new(X:Int, Y:Int, Width:Int, Height:Int, Zoom:Float = 0)
+	public function new(X:Int, Y:Int, Width:Int, Height:Int, Zoom:Float = 0, ZoomSpeed:Float = 25, ZoomMargin:Float = 0.25)
```
Gama11: I feel like these don't necessarily need to be in the constructor, having a lot of arguments always has an impact on readability.

Plus, your forgot to actually use the parameter values to initialize the variables. ;)

flixel/addons/display/FlxZoomCamera.hx
```diff
@@ -114,4 +114,18 @@ class FlxZoomCamera extends FlxCamera
 		flashSprite.scaleX = X;
 		flashSprite.scaleY = Y;
 	}
-}
\ No newline at end of file
+	/**
+	 * Given a value passed in, updates the speed of zooming. Useful for real time changes.
+	 */
+	public function setZoomSpeed(Value:Float):Void
```
Gama11: The variables should just be public instead of adding setter functions.

Gama11:
You're targetting the wrong branch, any changes go to dev.

flixel/addons/display/FlxZoomCamera.hx
AustinEast: Oops! I was copying from my local file and forgot to add that change in. I do feel at least the zoom speed is pretty useful to have there.

flixel/addons/display/FlxZoomCamera.hx
AustinEast: That would work too.

Gama11: Done in 44499cb.

PR #226, opened 2016-02-10
Optional Graphic for new backdrop

Tembac:
This is a really simple change. On FlxSprite the graphic is optional because we can load it later. I think the same must be done for backdrops because we can load the graphic later, as i'm doing on my game.


Gama11:
Have you checked that `draw()` can deal with a null graphic? I don't see any checks there.

Tembac:
I'm not sure why it shows 2 commits because I only made the changes from the second commit. Is this normal?

Gama11:
Are you sure `draw()` is fine with this in all cases? In the `FlxG.renderBlit` section, it accesses `graphic` without a null check for example.

Gama11:
Yeah, doesn't really look like this has been tested, crashes on both renderers.

Tembac:
Ups, Sorry. I though it was like that because I loaded the graphic immediately.
Thanks for fixing it!

PR #227, opened 2016-02-10
Update `draw()` of `FlxSkewedSprite`

adrianulima:
Tried to make it a `IFlxEffect`, but it works better as it is. Then I
just updated the `draw` method to override just `drawComplex` avoiding
some redundancy.


Gama11:
>20 additions and 65 deletions.

Nice! :+1:

Issue #228, opened 2016-02-19
Improve IFlxEffects


adrianulima:
This issue is to list next improvements on `IFlxEffect` that is planned, and if someone have a new idea for it:

FlxTrailEffect:
- [ ] Add support for angle/rotation;
- [ ] Add support for flip;
- [ ] Add support for scale;

FlxOutlineEffect:
- [ ] Add property `quality` similar to `FlxText#borderQuality`;
- [ ] Add `FlxOutlineMode` enum;
- [ ] Add `FlxOutlineMode.NORMAL` similar to `FlxTextBorderStyle.OUTLINE`;
- [ ] Add `FlxOutlineMode.FAST` similar to `FlxTextBorderStyle.OUTLINE_FAST`;

New IFlxEffects:
- [ ] Add `FlxShadowEffect` similar to `FlxTextBorderStyle.SHADOW`;

General:
- [ ] Try to reduce `BitmapData` new instances.

I think `IFlxEffect` is too young yet and have things to be improved, but when people start using it and help us it will be very important to many other cases.
`FlxCamera` has it own `shake` method that can be replaced by `FlxShakeEffect`;
`FlxText` has it own `border` that can be replaced by `FlxOutlineEffect` + `FlxShadowEffect`.

Gama11:
I'm not sure `FlxText` border styles should be replaced by `FlxEffectSprite` effects... You'd need an additional `FlxEffectSprite` instance just to get an outline effect.

adrianulima:
Yeah, what I'm saying is that `IFlxEffect` has potential to be more used in the future when we find all problems, implement all improvements and add more effects. Probably a far future `(5.0.0)`.

I thought about instead of a class`FlxEffectSprite` if we could implement all the logic in the bitmapData: `FlxBitmapData`.

xerosugar:
Suggestion for a new type of effect (or an "upgraded" version of FlxWaveEffect maybe?)
It works almost like the wave effect we have now, but the one I'm thinking adds distortion in a different way. In old 2D games the effect was often used in underwater scenes or to give the impression of heat (like in deserts).

Examples:
https://youtu.be/rXKyFZ-O6yE?t=51 (under water)
https://youtu.be/X8sXXNUc9GI?t=265 (under water, a bit more discrete)
https://youtu.be/69xfXl3rBoM?t=7 (desert)

[...]

PR #229, opened 2016-02-22
FlxEffects minor improvements

adrianulima:
FlxTrailEffect: Remove `new Matrix()` and `new ColorTransform()` from every frame
And some minor cleanup in the other effects.

I ignored `FlxOutlineEffect` for now because I'm rewriting it like described in: https://github.com/HaxeFlixel/flixel-addons/issues/228


adrianulima:
@Gama11 I have a refactored `FlxOutlineEffect` class, should I commit it here or open a new PR?

Gama11:
Separate would be nicer.

PR #232, opened 2016-02-28
FlxWaveEffect: add interlaceOffset

adrianulima:
Added `interlaced` param to `FlxWaveEffect`, I saw on Twitter and thought that can be useful.
Ref: [Twitter post](https://twitter.com/gamepopper/status/703687684737376256) and [video](https://www.youtube.com/watch?v=Yg2yQFN8hO4&feature=youtu.be&t=3m38s)

> This is my first PR with Mac, Sublime and SourceTree. Sorry if I did something wrong.

### DIFF
--- a/flixel/addons/effects/chainable/FlxWaveEffect.hx
+++ b/flixel/addons/effects/chainable/FlxWaveEffect.hx
@@ -20,15 +20,15 @@ class FlxWaveEffect implements IFlxEffect
 	public var offset(default, null):FlxPoint = FlxPoint.get();
 	
 	/**
-	 * Which mode we're using for the effect
+	 * Which mode we're using for the effect.
 	 */
 	public var mode:FlxWaveMode;
 	/**
-	 * How fast should the wave effect be (higher = faster)
+	 * How fast should the wave effect be (higher = faster).
 	 */
 	public var speed:Float;
 	/**
-	 * The 'center' of our sprite (where the wave effect should start/end)
+	 * The 'center' of our sprite (where the wave effect should start/end). Must be a value between 0 and 1;
 	 */
 	public var center:Float;
 	/**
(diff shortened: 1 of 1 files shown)


flixel/addons/effects/chainable/FlxWaveEffect.hx
```diff
@@ -71,13 +75,14 @@ class FlxWaveEffect implements IFlxEffect
 	 * @param	Wavelength	How long waves are.
 	 * @param	Direction	Which Direction you want the effect to be applied (HORIZONTAL or VERTICAL).
 	 */
-	public function new(?Mode:FlxWaveMode, Strength:Int = 10, Center:Float = -1, Speed:Float = 3, Wavelength:Int = 5, ?Direction:FlxWaveDirection)
+	public function new(?Mode:FlxWaveMode, Strength:Int = 10, Center:Float = -1, Speed:Float = 3, Wavelength:Int = 5, ?Direction:FlxWaveDirection, ?Interlaced:Bool = false)
```
Gama11: `?` shouldn't be used for basic types. This will implicitly turn this into a `Null<Bool>`.

http://stackoverflow.com/questions/31914614/is-question-mark-operator-in-haxe-really-necessary/31916466#31916466

adrianulima: Should I remove all `?`?
And can I change the params order to some like: `Strengh`, `Speed`, `Wavelength`, `Interlaced`, `Direction`, `Mode`, `center`?

Gama11: No, `?` is fine for non-basic types (for example the enum arguments).

Changing the sorting would break backwards compatibility.

adrianulima:
Done, removed the `?` from basic `Interlaced`. Thanks for teach this to me again.

Gama11:
This is a cool-looking feature, but I'm not sure a `Bool` and the hardcoded offset of `length * 0.5` is flexible enough. What if the user wants a different offset?

How about an `interlaceOffset:Float` which defaults to 0 and is a factor with a range of 0-1?

adrianulima:
Done, now the offset interlace ratio is working between 0 and 1.

flixel/addons/effects/chainable/FlxWaveEffect.hx
```diff
@@ -175,9 +173,21 @@ class FlxWaveEffect implements IFlxEffect
...
+		var waveOffset:Float = 0;
+		if (interlaceOffset != 0 && interlaceOffset != 1 && FlxMath.isOdd(p))
+		{
+			waveOffset = Math.PI * 2 * interlaceOffset;
+		}
+
+		return FlxMath.fastSin((p / wavelength) + _time + waveOffset);
+	}
+
+	private function set_interlaceOffset(InterlaceOffset:Float):Float
```
Gama11: Btw, this could in theory be written as a one-liner:

```haxe
return interlaceOffset = FlxMath.bound(InterlaceOffset, 0, 1);
```

PR #236, opened 2016-03-03
FlxEffectSprite: added `updateTargetAnimation`

adrianulima:
Needed when both effectSprite eh targetSprite are added to a state to avoid calling animation twice.


Gama11:
Do you think `updateTargetAnimation` would be better? `updateTarget` could be a bit misleading, it sounds like it calls `update()`.

adrianulima:
Done, also done the cleanup you suggested to `FlxWaveEffect`

PR #239, opened 2016-03-13
Flxcontrolhandler invert directions

winterismute:
Implements a simple way to invert movement of the character controlled by FlxControlHandler. Might be triggered by game effect f.e. to confuse the player.


flixel/addons/plugin/control/FlxControlHandler.hx
```diff
@@ -188,6 +188,9 @@ class FlxControlHandler
 	private var _clockwiseKey:String;
 	private var _thrustKey:String;
 	private var _reverseKey:String;
+	// Signal movement inverted on hor/vert axis
+	private var invertMovementHorizontally(get, null):Bool;
```
Gama11: `(get, null)` makes no sense for `private` properties.

Gama11:
I think instead of making these properties and having a `invertMovement()` function, these should be regular, public variables. Perhaps named `flipX` and `flipY` to match `FlxSprite` API?

Gama11:
"invert" may be more natural for movement than "flip" though, not sure...

winterismute:
@Gama11 well, I see that in the whole flixel there are different API conventions for different classes... But yes, maybe I should switch to having just "invertX" and "invertY" as public bools, in the end setters do not do any additional work... What do you think?

Gama11:
Sounds good.

winterismute:
@Gama11 This might be done now.

Gama11:
Thanks for the heads-up, there's no GitHub notifications for new commits on pull requests.

PR #243, opened 2016-03-28
FlxNapeSprite: add support for compound bodies

thegoodideaco:
If the body is part of a compound, then it's space cannot be changed, instead must be added/removed from the compound bodies list

### DIFF
--- a/flixel/addons/nape/FlxNapeSprite.hx
+++ b/flixel/addons/nape/FlxNapeSprite.hx
@@ -93,8 +93,7 @@ class FlxNapeSprite extends FlxSprite
 			updatePhysObjects();
 		}
 	}
-
-	/**
+	
 	/**
 	 * Handy function for "killing" game objects.
 	 * Default behavior is to flag them as nonexistent AND dead.
@@ -103,13 +102,12 @@ class FlxNapeSprite extends FlxSprite
 	{
 		super.kill();
 		
-		if (body != null && body.compound == null)
+		if (body != null)
 		{
-			body.space = null;
-		}
-		else if (body.compound != null)
-		{
-			body.compound.bodies.remove(body);
+			if (body.compound == null)
+				body.space = null;
+			else
+				body.compound.bodies.remove(body);
 		}
 	}
 	
@@ -121,13 +119,12 @@ class FlxNapeSprite extends FlxSprite
 	{
 		super.revive();
 		
-		if (body != null && body.compound == null)
+		if (body != null)
 		{
-			body.space = FlxNapeSpace.space;
-		}
-		else if (body.compound != null)
-		{
-			body.compound.bodies.add(body);
+			if (body.compound == null)
+				body.space = FlxNapeSpace.space;
+			else
+				body.compound.bodies.add(body);
 		}
 	}
 	
--- a/flixel/addons/nape/FlxNapeSprite.hx
+++ b/flixel/addons/nape/FlxNapeSprite.hx
@@ -95,6 +95,7 @@ class FlxNapeSprite extends FlxSprite
 	}
 
 	/**
+	/**
 	 * Handy function for "killing" game objects.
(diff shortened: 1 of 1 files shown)


flixel/addons/nape/FlxNapeSprite.hx
```diff
@@ -95,6 +95,7 @@ class FlxNapeSprite extends FlxSprite
 	}
 
 	/**
+	/**
```
Gama11: Seems the `/**` here was duplicated?

Gama11:
Any nape demo or code I could use to test this with?

thegoodideaco:
I can make a quick example if you like.

flixel/addons/nape/FlxNapeSprite.hx
thegoodideaco: ahh crap, fixing

flixel/addons/nape/FlxNapeSprite.hx
```diff
@@ -116,10 +121,14 @@ class FlxNapeSprite extends FlxSprite
 	{
 		super.revive();
 		
-		if (body != null)
+		if (body != null && body.compound == null)
```
Gama11: Are you sure it shouldn't look like this?

```haxe
if (body != null)
{
	if (body.compound == null)
		body.space = FlxNapeSpace.space;
	else
		body.compound.bodies.add(body);
}
````

Right now it would add `null` to `compound.bodies`.

**Comment by thegoodideaco** [comment]
@Gama11 : https://github.com/thegoodideaco/FlxNapeSpriteTest for an example.

**Comment by Gama11** [comment]
Thanks!

`set_physicsEnabled()` can still cause a crash. Should probably move the logic for adding a body / removing a body into separate methods to avoid redundancy and to be usable from there.

**Comment by Gama11** [comment]
Any update? :)

**Comment by thegoodideaco** [comment]
Should be ready to rock

File: flixel/addons/nape/FlxNapeSprite.hx [hunk 3]
```diff
@@ -217,8 +226,23 @@ class FlxNapeSprite extends FlxSprite
...
 		{
-			if (FlxNapeSpace.space != null)
-				FlxNapeSpace.space.bodies.remove(body);
+			
+			body.constraints.foreach(function(c:Constraint)
+			{
+				c.active = false;
+			});
+			
+			if (body.compound == null)
```
Gama11: This code section looks like an exact copy of the one in `revive()`.

flixel/addons/nape/FlxNapeSprite.hx
```diff
@@ -217,8 +226,23 @@ class FlxNapeSprite extends FlxSprite
 	{
 		if (body != null) 
 		{
-			if (FlxNapeSpace.space != null)
-				FlxNapeSpace.space.bodies.remove(body);
+			
+			body.constraints.foreach(function(c:Constraint)
```
Gama11: Why is this necessary?

Gama11:
The most recent commit doesn't seem to have any impact on the `set_physEnabled` crash:

```
Error: Cannot set the space of a Body belonging to a Compound, only the root Compound space can be set
```

(see https://github.com/Gama11/FlxNapeSpriteTest)

thegoodideaco:
This is odd, let me look

thegoodideaco:
i think we need to add an additional property to store any compound it might be apart of. If we disable physics, we have to remove it from the compound instead of from space. We need something to reference if they enable physics again

Gama11:
Hm... this is getting pretty complex. I should finally get around to setting up unit tests for flixel-addons so we can test all these different cases..

thegoodideaco:
I agree

Gama11:
Closing this for now since there's still some unresolved issues, I created an issue as a reminder: #321

PR #244, opened 2016-04-05
Added alpha support for FlxBackdrop

### LINKED ISSUE #241: FlxBackdrop doesn't support custom alpha
As the title says, even though you can set the alpha because it inherits from FlxSprite, it doesn't have any effect.

I already have code for both renderBlit and quadBatch render paths that I'm going to clean up. I'll submit a PR later.

diegosolo:
Fixes #241
Based on the work of  Ivan Vodopiviz (ivodopiviz)

### DIFF
--- a/flixel/addons/display/FlxBackdrop.hx
+++ b/flixel/addons/display/FlxBackdrop.hx
@@ -1,7 +1,6 @@
 package flixel.addons.display;
 
 import flash.display.BitmapData;
-import flash.geom.ColorTransform;
 import flash.geom.Point;
 import flash.geom.Rectangle;
 import flixel.FlxG;
--- a/flixel/addons/display/FlxBackdrop.hx
+++ b/flixel/addons/display/FlxBackdrop.hx
@@ -30,7 +30,6 @@ class FlxBackdrop extends FlxSprite
 	private var _spaceX:Int = 0;
 	private var _spaceY:Int = 0;
 	
-	private var _colorTransform:ColorTransform;
 	private var _blitPixels: BitmapData;
 	
 	/**
@@ -73,7 +72,6 @@ class FlxBackdrop extends FlxSprite
 		_spaceY = SpaceY;
 		
 		_ppoint = new Point();
-		_colorTransform = new ColorTransform();
 		
 		scrollFactor.x = ScrollX;
 		scrollFactor.y = ScrollY;
@@ -87,9 +85,9 @@ class FlxBackdrop extends FlxSprite
 	override public function set_alpha(alpha:Float):Float
 	{
 		super.alpha = alpha;
-		_colorTransform.alphaMultiplier = alpha;
+		colorTransform.alphaMultiplier = alpha;
 		_blitPixels.copyPixels(pixels, _flashRect2, _flashPointZero, null, null, true);
-		_blitPixels.colorTransform(_flashRect2, _colorTransform);
+		_blitPixels.colorTransform(_flashRect2, colorTransform);
 
 		return alpha;
 	}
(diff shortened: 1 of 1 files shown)


diegosolo:
Hi, I've seen you may need a project for testing, so I'm attaching one just in case.

What I did basically was to take the modifications made by Ivan and fix the lack of support for flash targets. For this I added a private BitmapData property ("_blitzPixels"), that I copied from the "pixels" property of FlxSprite, in order to apply a colorTransform to it when the alpha changed. Applying a colorTransform to the BitmapData object alters irretrievably the pixels, so it has to be restored each time a new transform needs to be added.

Then, the BitmapData is adjusted (width and height) in every regenGraphic call. The pixels are copied from the "pixels" property of FlxSprite and the colorTransform is applied, on each regenGraphic call and in every alpha change.

That's basically it...

[Backdrop.zip](https://github.com/HaxeFlixel/flixel-addons/files/205371/Backdrop.zip)

flixel/addons/display/FlxBackdrop.hx
```diff
@@ -29,6 +30,9 @@ class FlxBackdrop extends FlxSprite
 	private var _spaceX:Int = 0;
 	private var _spaceY:Int = 0;
 	
+	private var _colorTransform:ColorTransform;
```
Gama11: Couldn't we just use the `ColorTransform` instance of the parent class?

diegosolo: Apparently yes! I tested it with the project I uploaded in the previous message and another project I'm working on. No errors encountered.

Do I apply the changes?

diegosolo: Done!

diegosolo:
I've been thinking about it a little more. When you, for example, tween the alpha property of the Backdrop, these lines...

_blitPixels.copyPixels(pixels, _flashRect2, _flashPointZero, null, null, true);
_blitPixels.colorTransform(_flashRect2, colorTransform);

...are called every cycle. Are these methods fast enough (specially being backdrops usually composed by large images)?

Gama11:
Yes, that would definitely be heavy on performance.

diegosolo:
As I'm a little lost, I've been checking other classes like `FlxSprite` and `FlxTilemap` and they both use `BitmapData` buffers to handle color transforms. For example, in `FlxSprite`, when a new alpha value is set, the `dirty` flag is set to `true` and the method `updateFramePixels` is called in the next `draw`. This method updates the `framePixels` (`BitmapData`) property and applies, if necessary, the color transform to it. So, basically, you're doing this...
```
framePixels = _frame.paint(framePixels, _flashPointZero, false, true);
framePixels.colorTransform(_flashRect, colorTransform);
```
...every time the alpha value changes. Now, paint ends up calling `copyPixels` in a `BitmapData` object, so there's not much difference with the proposed implementation for `FlxBackdrop`... I think. Would you confirm?

PD: Yeah, we could polish the code to prevent a few unnecessary calls to `copyPixels`, but not much more...

diegosolo:
It's fixed! The problem with `FlxBackdrop` is that it overrides completely the `draw` method in its parent. Then, the `dirty` flag is never checked, the `updateFramePixels` never called and the `framePixels` never updated. So I removed all changes and simply added (like in `FlxSprite`) the lines...
```
if (dirty)
	calcFrame(useFramePixels);
```
...befor the lines...
```
_flashRect2.setTo(0, 0, graphic.width, graphic.height);
camera.copyPixels(frame, framePixels, _flashRect2, _ppoint);
```
... and voila! It works like a charm. I'll update the request...

diegosolo:
Hi! Any reason why this hasn't been merged? Is there anything else I can do?

Gama11:
Wow, this PR got a lot simpler since I last checked it. :+1: Seems to be working fine as well.

Thanks for the contribution, and sorry for the delay!

diegosolo:
Thanks! This is my first collaboration to the Open Source community so I'm pretty happy!

Just a side note. In native targets, the backgrounds overlap something around one pixel. This is pretty visible when the alpha value goes away from 1 or 0. I tried setting `useScaleHack` (which I don't really know what is for) to `false` with no luck. I may try to work on it in the future if I'm not so busy. I just wanted to let you know...

diegosolo:
Sorry, my bad. It actually was the `useScaleHack` variable after all. Changing it to `false` solves the problem. Cheers!!

PR #245, opened 2016-04-07
Tiled: added rootPath string so that you can properly parse complex T…

larsiusprime:
…iled maps, added csv parsing to TiledTileLayer.get_tileArray()

### DIFF
--- a/flixel/addons/editors/tiled/TiledMap.hx
+++ b/flixel/addons/editors/tiled/TiledMap.hx
@@ -49,7 +49,7 @@ class TiledMap
 	
 	/**
 	 * @param data Either a string or XML object containing the Tiled map data
-	 * @param homePath Path to use as root to resolve any internal file references
+	 * @param rootPath Path to use as root to resolve any internal file references
 	 */
 	public function new(data:FlxTiledMapAsset, rootPath:String="")
 	{
--- a/flixel/addons/editors/tiled/TiledMap.hx
+++ b/flixel/addons/editors/tiled/TiledMap.hx
@@ -51,7 +51,7 @@ class TiledMap
 	 * @param data Either a string or XML object containing the Tiled map data
 	 * @param homePath Path to use as root to resolve any internal file references
 	 */
-	public function new(data:FlxTiledAsset, rootPath:String="")
+	public function new(data:FlxTiledMapAsset, rootPath:String="")
 	{
 		var source:Fast = null;
 		var node:Fast = null;
@@ -179,4 +179,4 @@ class TiledMap
 	}
 }
 
-typedef FlxTiledAsset = OneOfTwo<String, Xml>;
+typedef FlxTiledMapAsset = OneOfTwo<String, Xml>;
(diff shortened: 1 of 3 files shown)


larsiusprime:
Context: an artist delivered a Tiled map to me that I couldn't properly parse with the existing version because it relied on internal file references, which seems like a complex use case that the base parser doesn't seem to account for. This makes it so that it can properly resolve those internal file references.

Also, I added the csv parsing to get_tileArray(), because I don't think it will always be clear to the user what format tiled is representing the internal tile arrays in, and thus know they should use FlxTIlemaps "loadFromCSV" instead -- this causes it to just give you your tile array regardless of what encoding it's using. This happened to me as get_tileArray() just returned nothing, and it wasn't until I went poking around in the file format that I discovered it was because the encoding was "csv" and that wasn't implemented in get_tileArray().

flixel/addons/editors/tiled/TiledTileSet.hx
```diff
@@ -30,7 +31,7 @@ class TiledTileSet
 	
 	public var tileImagesSources:Array<TiledImageTile>;
 	
-	public function new(data:Dynamic)
+	public function new(data:Dynamic, rootPath:String="")
```
Gama11: Hm, we should really use one of the `OneOf` types to clearly define the acceptable types here like we do for `FlxGraphicAsset` etc..

Gama11:
Got a file to test these changes with?

larsiusprime:
@Gama11 just sent you on slack

flixel/addons/editors/tiled/TiledTileSet.hx
larsiusprime: Do you want me to go ahead and do that then?

Gama11: Feel free to.

larsiusprime:
@Gama11: Okay, added it. There was already a OneOfTwo with the name "FlxTiledAsset" that didn't seem like the right thing here, so I gave it a clearer name... would that technically be a breaking change? I can revert that.

Also I put in a slightly cleaner way to deal with the legacy Byte classes, but your call on that one too.

flixel/addons/editors/tiled/TiledMap.hx
```diff
@@ -45,14 +45,19 @@ class TiledMap
 	private var noLoadHash:Map<String, Bool> = new Map<String, Bool>();
 	private var layerMap:Map<String, TiledLayer> = new Map<String, TiledLayer>();
 	
+	private var rootPath:String="";
+	
 	/**
 	 * @param data Either a string or XML object containing the Tiled map data
+	 * @param homePath Path to use as root to resolve any internal file references
```
Gama11: `homePath` in the javadoc here seems incorrect?

flixel/addons/editors/tiled/TiledTileSet.hx
```diff
@@ -200,3 +211,6 @@ class TiledTileSet
 		return new Rectangle((ID % numCols) * tileWidth, (ID / numCols) * tileHeight);
 	}
 }
+
+typedef ValidByteArray = #if (lime_legacy || openfl <= "3.4.0") ByteArray #else ByteArrayData #end;
```
Gama11: Should be `private`.

larsiusprime:
@Gama11  anything else?

PR #246, opened 2016-04-14
Added FlxMosaic

xerosugar:


Gama11:
I'm not convinced it's necessary to have this in flixel-addons - it's just a really simple wrapper around `Shader`. I prefer to keep it in the demo like you already did in https://github.com/HaxeFlixel/flixel-demos/pull/225.

PR #247, opened 2016-04-25
Tiled objects should default to having a name

JoeCreates:
I had carried this over from the old tiled stuff, but have since realised that it does not make sense to give objects a default name. With the default name like this, in order to check if an object did not have a name, you must check for the name "[object]", which is not intuitive. Checking for an empty string is more intuitive, and better reflects the empty field in the editor.

### DIFF
--- a/flixel/addons/editors/tiled/TiledObject.hx
+++ b/flixel/addons/editors/tiled/TiledObject.hx
@@ -73,7 +73,7 @@ class TiledObject
 	{
 		xmlData = source;
 		layer = parent;
-		name = (source.has.name) ? source.att.name : "[object]";
+		name = (source.has.name) ? source.att.name : "";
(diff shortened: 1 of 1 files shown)


Gama11:
Why is an empty string better than null?

JoeCreates:
I'm mainly concerned that null was previously not possible (well, unless somebody explicitly set it to null), so there is presently likely a lot of code that presumes the name is not null and would then give a null reference error.

I'd change it to null if you think that would be better, though the other properties should probably also be changed to default to null in that case. I'm not sure there is any benefit to doing that, though, seeing as in tiled it is not possible to remove the name or type field of an object. There is no way to distinguish, on input, between a non-existent property and an empty field; you're only option is to leave those fields blank.

PR #251, opened 2016-05-15
Add offset support for Tiled.

starry-abyss:
I'm not sure what meaning do properties x and y in derived classes have, but this is how the positioning of layers stored in .tmx file. At least in Tiled 0.16.1

### DIFF
--- a/flixel/addons/editors/tiled/TiledLayer.hx
+++ b/flixel/addons/editors/tiled/TiledLayer.hx
@@ -18,6 +18,8 @@ class TiledLayer
 	public var opacity:Float;
 	public var visible:Bool;
 	public var properties:TiledPropertySet;
+	public var offsetx:Float;
+	public var offsety:Float;
 
 	private function new(source:Fast, parent:TiledMap)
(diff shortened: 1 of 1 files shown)


flixel/addons/editors/tiled/TiledLayer.hx
```diff
@@ -18,6 +18,8 @@ class TiledLayer
 	public var opacity:Float;
 	public var visible:Bool;
 	public var properties:TiledPropertySet;
+	public var offsetx:Float;
```
Gama11: Please use proper camel case (`offsetX`).

Gama11: Another option might be using a `FlxPoint` object, that'd make it consistent with `FlxSprite`'s `offset`...

starry-abyss:
@Gama11 I wanted to be consistent with this addon :-) FlxPoint is only used there for collision polygon vertices, as far as I see

Gama11:
:+1:

PR #254, opened 2016-05-23
Allow FlxBullet offset to count for parent's angle

alihassan0:
to easily test just use this project [flxWeaponTest.zip](https://github.com/HaxeFlixel/flixel-addons/files/278162/flxWeaponTest.zip)


Gama11:
The diff here only shows the changes from your other pull request #257. Could you reopen this one with the correct changes?

PR #257, opened 2016-06-20
FlxFSM: fix an invalid field access error on Neko

alihassan0:
`map.exists(null);` breaks the application on neko .. 
so as a workaround I check if the parameters is not null first


Gama11:
Looks like you accidentally added the changes from #254 as well here?

alihassan0:
I branch before I fixed this .. but i think I forget that means all my past commits are there too 
should I close this and create a new one ?

Gama11:
You could also revert that commit.

Gama11:
Thanks!

PR #258, opened 2016-07-27
Add FlxPosition

DleanJeans:
FlxPosition provides a pack of util functions for positioning FlxObjects through static extesion (using).
It also comes with screen position variables (actually getters) like screen center, top right, bottom right, bottom left.

If you need positioning for non-FlxObjects then there's FlxDynamicPosition (For some reasons, FlxDynamicPosition doesn't work well on FlxObjects so there has to be two different classes)

~~There's also a modified FlxText to be compatible with FlxPosition since FlxText width and height properties don't return real width and height the last time I used it.~~ Not needed anymore.


flixel/addons/util/position/FlxDynamicPosition.hx
```diff
@@ -0,0 +1,215 @@
...
+	public static var screenTopLeft(get, null):FlxPoint;
+	public static var screenMidTop(get, null):FlxPoint;
+	public static var screenTopRight(get, null):FlxPoint;
+	public static var screenMidRight(get, null):FlxPoint;
+	public static var screenBottomRight(get, null):FlxPoint;
+	public static var screenMidBottom(get, null):FlxPoint;
+	public static var screenBottomLeft(get, null):FlxPoint;
+	public static var screenMidLeft(get, null):FlxPoint;
+	
+	public static inline function getTopLeft(object:Dynamic):FlxPoint {
```
Gama11: I don't think we should promote the use of `Dynamic` in this way, it's bad practice and bad for performance.

Gama11:
Although this looks useful, I'd prefer not to add this flixel-addons, since we already have a similar feature with flixel-ui's [`Anchor`](https://github.com/HaxeFlixel/flixel-ui/blob/dev/flixel/addons/ui/Anchor.hx).

If you find anything from here missing, perhaps it could be added there? :)

DleanJeans:
[`Anchor`](https://github.com/HaxeFlixel/flixel-ui/blob/dev/flixel/addons/ui/Anchor.hx) seems pretty complicated. To use it, you have to create a new object by passing a bunch of arguments.
With FlxPosition, you only need to import it by static extension and type this:

```Haxe
// This will move object1 right next to the right of object2
// and align the center-y of object1 to center-y of object2

object1.setMidLeft(object2.getMidRight());
```

Gama11:
Could add some default values to some of the constructor arguments I guess... Feel free to suggest improvements!

PR #261, opened 2016-08-21
Allow FlxBullet offset to count for parent's angle

alihassan0:
to easily test just use this project [flxWeaponTest.zip](https://github.com/HaxeFlixel/flixel-addons/files/278162/flxWeaponTest.zip)

### DIFF
--- a/flixel/addons/weapon/FlxWeapon.hx
+++ b/flixel/addons/weapon/FlxWeapon.hx
@@ -513,7 +513,7 @@ class FlxTypedWeapon<TBullet:FlxBullet>
 
 enum FlxWeaponFireFrom
 {
-	PARENT(parent:FlxSprite, offset:FlxBounds<FlxPoint>, ?useParentAngle:Bool);
+	PARENT(parent:FlxSprite, offset:FlxBounds<FlxPoint>, useParentAngle:Bool = false);
 	POSITION(position:FlxBounds<FlxPoint>);
 }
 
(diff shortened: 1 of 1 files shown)


flixel/addons/weapon/FlxWeapon.hx
```diff
@@ -479,7 +513,7 @@ class FlxTypedWeapon<TBullet:FlxBullet>
 
 enum FlxWeaponFireFrom
 {
-	PARENT(parent:FlxSprite, offset:FlxBounds<FlxPoint>);
+	PARENT(parent:FlxSprite, offset:FlxBounds<FlxPoint>, ?useParentAngle:Bool);
```
Gama11: Can you replace `?useParentAngle:Bool` with `useParentAngle:Bool = false`? There's no good reason to turn this into a `Null<Bool>` (see http://stackoverflow.com/questions/31914614/is-question-mark-operator-in-haxe-really-necessary/31916466#31916466).

alihassan0: Done :+1:

Gama11:
Thanks! The test project you provided was quite helpful for reviewing the changes. :+1:

Gama11:
Hm, nevermind about my earlier comment, I could've sworn that compiled when I tried it - seems like Haxe simply doesn't allow default values for enum constructors like it does with functions.s

PR #263, opened 2016-08-30
Add scale argument to FlxPexParser.

bendmorris:
Adds a `scale` argument to FlxPexParser.parse which will scale particle size and positions for the resulting emitter to effectively zoom in/out on the emitter. Also an unrelated bug fix on line 58.

It might be nice if FlxEmitters themselves had the ability to scale like this so you could just scale the resulting emitter, but AFAIK they don't currently, you'd have to modify position/speed/scale etc. on your own.

### DIFF
--- a/flixel/addons/editors/pex/FlxPexParser.hx
+++ b/flixel/addons/editors/pex/FlxPexParser.hx
@@ -25,9 +25,10 @@ class FlxPexParser
 	 * @param	data			The data to be parsed. It has to be an ID to the assets file, a file embedded with @:file(), a string with the content of the file or a XML object.
 	 * @param	particleGraphic	The particle graphic
 	 * @param	emitter			(optional) A FlxEmitter. Most properties will be overwritten!
+	 * @param	scale			(optional) Used to scale the resulting emitter. Will scale both the texture size and positions of resulting particles.
 	 * @return	A new emitter	
 	 */
-	public static function parse<T:FlxEmitter>(data:Dynamic, particleGraphic:FlxGraphicAsset, ?emitter:T):T
+	public static function parse<T:FlxEmitter>(data:Dynamic, particleGraphic:FlxGraphicAsset, ?emitter:T, scale:Float = 1):T
 	{
 		if (emitter == null)
 		{
@@ -54,7 +55,7 @@ class FlxPexParser
 		var angle = minMax("angle", config);
 		
 		var startSize = minMax("startParticleSize", config);
-		var finishSize = minMax("finishParticleSize", "FinishParticleSizeVariance", config);
+		var finishSize = minMax("finishParticleSize", "finishParticleSizeVariance", config);
 		var rotationStart = minMax("rotationStart", config);
 		var rotationEnd = minMax("rotationEnd", config);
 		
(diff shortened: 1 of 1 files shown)


flixel/addons/editors/pex/FlxPexParser.hx
```diff
@@ -25,9 +25,10 @@ class FlxPexParser
 	 * @param	data			The data to be parsed. It has to be an ID to the assets file, a file embedded with @:file(), a string with the content of the file or a XML object.
 	 * @param	particleGraphic	The particle graphic
 	 * @param	emitter			(optional) A FlxEmitter. Most properties will be overwritten!
+	 * @param	scale			(optional) Used to scale the resulting emitter. Will scale both the texture size and positions of resulting particles.
 	 * @return	A new emitter	
 	 */
-	public static function parse<T:FlxEmitter>(data:Dynamic, particleGraphic:FlxGraphicAsset, ?emitter:T):T
+	public static function parse<T:FlxEmitter>(data:Dynamic, particleGraphic:FlxGraphicAsset, ?emitter:T, ?scale:Float=1):T
```
Gama11: I don't think `scale` should have a `?`, that unnecessarily makes it a `Null<Float>` (see my SO answer [here](http://stackoverflow.com/questions/31914614/is-question-mark-operator-in-haxe-really-necessary/31916466#31916466)). Or am I missing something?

Also, since you're already at it: Flixel's code style has spaces before and after the `=`.

Gama11:
Just to clairfy I'm understanding this correctly: `scale` isn't something that's built into the pex parser, this is something that you added because it's useful?

Either way, the `FlxPexParser` demo should probably be updated to showcase the new functionality in some way.

bendmorris:
You're right about scale, fixed that.

And yes, this isn't a feature of pex per se, just for convenience. I'm editing particle effects with Particle Designer and want to be able to tweak the sizes easily at runtime from a single pex file. FlxSpine provides a similar scale arg to scale skeletons when parsing spine files.

Gama11:
Should I wait for an update of the demo before merging this or not?

flixel/addons/editors/pex/FlxPexParser.hx
```diff
@@ -54,7 +55,7 @@ class FlxPexParser
 		var angle = minMax("angle", config);
 		
 		var startSize = minMax("startParticleSize", config);
-		var finishSize = minMax("finishParticleSize", "FinishParticleSizeVariance", config);
+		var finishSize = minMax("finishParticleSize", "finishParticleSizeVariance", config);
```
Gama11: Are you sure about this? In the `FlxPexParser` demo, the `.pex` file actually uses `FinishParticleSizeVariance`, and changing this breaks it:

```
particleEmitterConfig is missing element finishParticleSizeVariance
	at haxe.xml._Fast::NodeAccess/resolve()
	at flixel.addons.editors.pex::FlxPexParser$/minMax()
	at flixel.addons.editors.pex::FlxPexParser$/parse()
```

Gama11: When I export the default particle effect on http://onebyonedesign.com/flash/particleeditor/, it's also capitalized for me.

bendmorris: Seems like a bug in that editor - Particle Designer doesn't capitalize (and no other node is capitalized.) I'll modify this to look for either version.

Gama11: Oh, I didn't realize there are multiple tools that produce this xml format. Could you add a link to that tool in the class docs, where we link to http://onebyonedesign.com/flash/particleeditor/?

bendmorris: Sure. It is apparently a typo which originated in PD and was eventually fixed: http://forum.starling-framework.org/topic/this-might-be-a-typo-in-staring-extension-particle-system so best to support either version.

Gama11: Agreed!

Gama11:
Btw, just fyi, I found some unused vars in `FlxPexParser`: https://github.com/HaxeFlixel/flixel-addons/issues/264

flixel/addons/editors/pex/FlxPexParser.hx
```diff
@@ -149,7 +150,14 @@ class FlxPexParser
 			maxColor: FlxColor.fromRGBFloat(minR + varR, minG + varG, minB + varB, minA + varA)
 		};
 	}
-	
+
+	private static inline function getNodeName(property:String, config:Fast):String
+	{
+		// for backwards compatibility, check for versions of properties that
```
Gama11: Oh, so this is an issue for all properties? I thought it was specifically `"finishParticleSizeVariance"`.

bendmorris: That's one I know of but there were apparently multiple typos over time: https://github.com/Gamua/Starling-Extension-Particle-System/blob/master/src/starling/extensions/PDParticleSystem.as#L260

It seems like it wouldn't hurt to check all nodes. If Fast had a case-insensitive way to resolve nodes that would be even better.

Gama11: Gotcha.. can't really hurt I guess! :)

Gama11:
Thanks!

flixel/addons/editors/pex/FlxPexParser.hx
Gama11: Looks like you forgot to add the link.. is it this tool? https://71squared.com/particledesigner

bendmorris: Yep.

PR #265, opened 2016-09-07
Update FlixelTextureLoader to use FlxAssets.

bendmorris:
Currently FlixelTextureLoader uses openfl.Assets.getBitmapData which may load a cached version of a previously loaded BitmapData. Using FlxAssets to bypass the cache, so after the FlxSpine is destroyed at state change a new version of the bitmap will be loaded.

### DIFF
--- a/flixel/addons/editors/spine/texture/FlixelTextureLoader.hx
+++ b/flixel/addons/editors/spine/texture/FlixelTextureLoader.hx
@@ -1,7 +1,7 @@
 package flixel.addons.editors.spine.texture;
 
-import openfl.Assets;
 import openfl.display.BitmapData;
+import flixel.system.FlxAssets;
 import spinehaxe.atlas.AtlasPage;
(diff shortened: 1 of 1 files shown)


Gama11:
`FlxAssets.getBitmapData()` ends up calling `openfl.Assets.getBitmapData()` as well... How does that "bypass" anything?

Gama11:
Could you provide a more concrete example of what issue this fixes?

bendmorris:
Flixel's internal cache is cleared on state change. OpenFL's isn't, so any active FlxSprite is destroyed, but the destroyed image is still sitting in the OpenFL cache and will be fetched next time you load the same asset. You can reproduce this easily:

- change FlxAssets to use "true" instead of "false" for useCache
- load a graphic in one state
- change to another state which uses the same graphic - the graphic is now invisible

Gama11:
Thanks for the explanation, that makes a lot of sense. :)

PR #267, opened 2016-09-12
FlxTypeText: allow adding / removing more than one char per frame

bendmorris:
With a low delay, the current implementation is only capable of showing or erasing one character per frame. This change allows using any delay, independent of the update rate.


Gama11:
Thanks!

PR #280, opened 2016-11-26
[Incomplete] Simplex noise, closes #266

### LINKED ISSUE #266: Simplex noise
I think we agreed on putting a Simplex noise generator in addons, but where in the package structure should it go?

I still need to clean it up a bit before submitting.

MSGhero:
Closes #266 

Right now, you get the noise value at a single x,y. I believe the tiling and octaves ones are how BitmapData's `perlinNoise` works.

Besides documentation, what else would be good to have or change, especially related to flixel? I should change tiles to allow for scale and persistence. Also need to check if this edited version compiles...

I tested the inlines a while back, and the functions were significantly faster with both of them in.


Gama11:
What's the status here?

MSGhero:
Status is I forgot this existed lol. Is it fine as is, or should there be more added?

MSGhero:
Added comments.

MSGhero:
Wait I have merge access? What

gamedevsam:
With great power comes great responsibility :-P

MSGhero:
Uhh uh ok. I think I need to add more `Flx` prefixes and then revisit it once more.

Gama11:
I'm not sure about starting a whole new `algorithms` package with this as its only contents. Would `flixel.addons.util` work?

Also, always hesitant about merging stuff that isn't used in a demo (the stuff without demos is usually more poorly maintained / tested / etc).

MSGhero:
Util is fine, and I thought up a demo already that I'll add as well.

MSGhero:
I was actually planning on finally doing this demo over this and next weekend since I'll be traveling

PR #284, opened 2016-12-24
Small memory optimizations (mostly involving use of haxe.ds.Vector instead of Array)

JesseTG:
No public APIs were changed; I just went around and changed some uses of `Array<T>` to `haxe.ds.Vector<T>` when the size was fixed and known before the construction of the `Vector`.  I didn't make any radical changes either, just some low-hanging fruit.  The most notable thing I did was in `FlxStarfield`; I removed the `FlxStar` helper class entirely and replaced it with five `Vector<Float>`'s, so that it can benefit from sequential cache use.  That way, we don't have to worry about the memory allocator placing `FlxStar`s wherever it pleases.

### DIFF
--- a/flixel/addons/display/FlxStarField.hx
+++ b/flixel/addons/display/FlxStarField.hx
@@ -106,11 +106,15 @@ private class FlxStarField extends FlxSprite
 {
 	public var bgColor:Int = FlxColor.BLACK;
 
+	// Why five Vectors instead of one that has an FlxStar class?  This is better
+	// for the cache, as all the data is being accessed sequentially so it'll be a
+	// little faster.
 	private var _starsX:Vector<Float>;
 	private var _starsY:Vector<Float>;
 	private var _starsD:Vector<Float>;
 	private var _starsR:Vector<Float>;
 	private var _starsSpeed:Vector<Float>;
+	// TODO: Can we speed this up more by just using one big Vector that's five times as long as StarAmount?
 
 	private var _depthColors:Array<Int>;
 	private var _minSpeed:Float;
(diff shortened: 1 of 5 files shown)


flixel/addons/display/FlxStarField.hx
```diff
@@ -106,8 +105,17 @@ class FlxStarField3D extends FlxStarField
 private class FlxStarField extends FlxSprite
 {
 	public var bgColor:Int = FlxColor.BLACK;
-	
-	private var _stars:Array<FlxStar>;
+
+	// Why five Vectors instead of one that has an FlxStar class?  This is better
+	// for the cache, as all the data is being accessed sequentially so it'll be a
+	// little faster.
+	private var _starsX:Vector<Float>;
```
Gama11: Hm... this might be faster, but the old version was definitely more readable.

Have you done any benchmarks to see what difference it actually makes in practice?

JesseTG: Nothing formal, I just ran Flixius twice (before and after the changes) and looked at the update times.  What would you suggest I do for a more formal benchmark?

Gama11: What was the difference you observed there?

JesseTG: Nothing remarkable (less than a millisecond improvement, if I remember correctly), but that's not considering the GC impact (which I did not evaluate).

Gama11:
There's too many merge conflicts here by now, and I still prefer the old version of `FlxStarField` when it comes to readability.

PR #288, opened 2017-02-06
TiledObject: fix gid handling with flipped objects, closes #287

### LINKED ISSUE #287: Help working with flipped objects on tiled
Flixel version: 4.2.0
Flixel addons version: 2.2.0
OpenFL version: 3.6.1
Lime version: 2.9.1
_____

I'm working with tiled for the levels and I want to use the Flipping variable to know if an object is flipped. I'm not sure how it works because that variable seems to change the Gid of the object. 
I'm using the gid to know which object is each one but when I flip it changes.

For example, an object without flip can have a gid that is: 4372 and the same object flipped have the gid: 2147488019

On TiledObject.hx I have functions to know if an object is flipped but I want to also know the gid before the flip so I can know which object is. 

I'm not sure if this is something that I'm supposed to handle on my code or maybe there is something I'm missing.

This can be tested on the tiled demo: https://github.com/HaxeFlixel/flixel-demos/tree/dev/Editors/TiledEditor

Tembac:
Fixed support for flipping tile objects from tiled. It is only working for flash target. I need help to solve it for other targets.

It is based on tiled guide about flipping: http://doc.mapeditor.org/reference/tmx-map-format/#tile-flipping

issue: #287


flixel/addons/editors/tiled/TiledObject.hx
```diff
@@ -93,6 +93,13 @@ class TiledObject
 		{
 			gid = Std.parseInt(source.att.gid);
 			
+			flippedHorizontally = (gid & FLIPPED_HORIZONTALLY_FLAG) < 0;
+			
+			flippedVertically = (gid & FLIPPED_VERTICALLY_FLAG) > 0;
+			
+			gid &= ~(FLIPPED_HORIZONTALLY_FLAG |
+                        FLIPPED_VERTICALLY_FLAG );
```
Gama11: This line seems to use spaces for indent.

flixel/addons/editors/tiled/TiledObject.hx
```diff
@@ -59,11 +59,11 @@ class TiledObject
 	/**
 	 * Whether the object is flipped horizontally.
 	 */
-	public var flippedHorizontally(get, null):Bool;
+	public var flippedHorizontally:Bool;
 	/**
 	 * Whether the object is flipped vertically.
 	 */
-	public var flippedVertically(get, null):Bool;
+	public var flippedVertically:Bool;
```
Gama11: `null`-set access means setting the property is only allowed inside this class. Is there a reason to lift that restriction?

Otherwise, it should be `(default, null)`.

Tembac:
@Gama11 is it better like this?

Do you know why it can be that it works for flash but not for windows?

flixel/addons/editors/tiled/TiledObject.hx
```diff
@@ -93,6 +93,16 @@ class TiledObject
 		{
 			gid = Std.parseInt(source.att.gid);
 			
+			flippedHorizontally = (gid & FLIPPED_HORIZONTALLY_FLAG) < 0;
+			trace("hor flip " + flippedHorizontally);
```
Gama11: Leftover debug traces?

flixel/addons/editors/tiled/TiledObject.hx
```diff
@@ -93,6 +93,16 @@ class TiledObject
...
+			flippedVertically = (gid & FLIPPED_VERTICALLY_FLAG) > 0;
+			trace("vert flip " + flippedVertically);
+
+			
+			gid &= ~(FLIPPED_HORIZONTALLY_FLAG |
+                        FLIPPED_VERTICALLY_FLAG );
```
Gama11: Still using space indentation here it look like.

Gama11:
>Do you know why it can be that it works for flash but not for windows?

Not yet, I have to check it out and give it a try locally.

Gama11:
In #287, you say the issue can be observed in the TiledEditor demo. However, I see the same traces there for both Flash, Cpp and Neko. Since you mentioned the fix only works on Flash, I assumed I should be seeing different results. Or do I need different traces?

If the TiledEditor demo currently doesn't "test" this bug (or at least it doesn't make a difference visually), can you adjust it so that it does?

alihassan0:
I just tested that brand on Neko,Linux mint 64bit.

Tembac:
I added a PR with the demo. It is strange because here the traces are different. 

The demo adds a flipped image and it crashes when it tried to load it on windows or neko.

Gama11:
The problem is that the original gid value of `3221225704` exceeds the valid range of a signed integer (which is `-2,147,483,648` to `2,147,483,647`). That's why the example code you linked used an `unsigned` int.

That's also why this doesn't compile:

```haxe
var i:Int = 3221225704;
```

Gama11:
I pushed a fix to your branch - works ok now as far as I can tell.

Note that just using `haxe.Int` isn't enough, you have to avoid `Std.parseInt()` as well, since the int it returns is already truncated to the Int32 range.

Tembac:
You are amazing @Gama11 ! 

Thanks for your help! It works great.

Will this be merged to flixel-addons?

Gama11:
Sure, was just waiting for your confirmation. :)

Gama11:
Urgh, good thing we have Travis, looks like `haxe.Int64` didn't have some of this API in Haxe 3.2.x... :/

Gama11:
Damn, `Int64.parseString()` wasn't available before then. Not sure we really want to break Haxe 3.2 compatiblity yet.

PR #289, opened 2017-02-08
Flipping object fix for horizontal

Tembac:
Fix for horizontal flip. Before I only checked for vertical. Sorry about this @Gama11


Gama11:
Thanks, should've looked at that code more closely. :)

PR #292, opened 2017-03-02
Add angle offsets to FlxWeapon

alihassan0:
it's useful when using `useParentAngle` since it allows to offset the bullet angle since you have no control on it.
For example if you have the bullet sprite directed upwards you can set the offset to 90 so that it looks right

### DIFF
--- a/flixel/addons/weapon/FlxWeapon.hx
+++ b/flixel/addons/weapon/FlxWeapon.hx
@@ -1,10 +1,10 @@
 package flixel.addons.weapon;
 
-import flixel.addons.weapon.FlxWeapon.FlxTypedWeapon;
 import flixel.FlxBasic;
 import flixel.FlxG;
 import flixel.FlxObject;
 import flixel.FlxSprite;
+import flixel.addons.weapon.FlxWeapon.FlxTypedWeapon;
 import flixel.group.FlxGroup.FlxTypedGroup;
 import flixel.input.touch.FlxTouch;
 import flixel.math.FlxAngle;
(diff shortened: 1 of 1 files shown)


Gama11:
API-wise, wouldn't it be better if this was another argument for the `PARENT` enum constructor (since it's only used in that case)?

alihassan0:
I thought of that .. but I was worried that this is making the instructor much longer. but if that is not a problem then you are right .. 

Also BTW .. the `FlxWeapon` Constructor docs is outdated .. if you want I can change it as well.

Gama11:
>making the instructor much longer

I'm not sure what you mean by that.

>Constructor docs is outdated .. if you want I can change it as well.

Yes, please. :)

alihassan0:
> I'm not sure what you mean by that.
sorry i meant the enum constructor

Gama11:
Oh, you mean you were concerned about there being too many parameters. :)

I think 4 isn't too much, especially if 2 of them are optional. Let's change it to an enum contructor argument.

Gama11:
Any update? :)

alihassan0:
Sorry for the delay .. I was just messing a lot with that  file and didn't had time to clean it. but I will do it now 
 
I also made a change in line [269](https://github.com/alihassan0/flixel-addons/blob/6fd39643c33ce48718442a226c673a27eae11b00/flixel/addons/weapon/FlxWeapon.hx#L223) should I make a separate PR for it .. or can I include it in this PR ?
 

```hx
actualOffset.subtract(parent.offset.x, parent.offset.y);
```

Gama11:
Seems like a bit of a different thing to me?

alihassan0:
sort of  .. this PR originally allows to add angle offset when using the parent angle .
The new line fixes a problem where if the parent sprite is offsetted then the origin will be offsetted which result in wrong placement. so it's just subtracts it first.

Gama11:
Ah, so it's necessary to make the change in this PR work right? Feel free to include it in that case.

PR #293, opened 2017-03-12
FlxTypeText: always set callbacks

winterismute:
When a complete/erase callback is set, it is basically impossible to unset it since the most obvious way (not to set anything when "start" is called again, for example) does not work: the code checks for nullity of the callback argument and in that case does not (re)set it. I think it must be simply always set.
Possibly, this might be a breaking change, if code calls "start" after an initial "start" with no callback, and expects to have the same callback. However, the version in this branch is the most logical imho.
I can write tests for this if needed.

### DIFF
--- a/flixel/addons/text/FlxTypeText.hx
+++ b/flixel/addons/text/FlxTypeText.hx
@@ -186,10 +186,7 @@ class FlxTypeText extends FlxText
 			skipKeys = SkipKeys;
 		}
 		
-		if (Callback != null)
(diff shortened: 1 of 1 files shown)


Gama11:
Agreed, this behavior seems a bit counter-intuitive (especially considering it wasn't really documented).

>it is basically impossible to unset it

Actually, the callbacks are assigned to public vars, you can just set those to null if you want. :)

PR #295, opened 2017-05-12
FlxTypeText: stop sounds when typing is complete

winterismute:
Inside "onComplete" (when typing has finished), stop all the sounds that are playing.
Had this issue in one of my prototypes. Can add unit test if you feel like it's needed.

### DIFF
--- a/flixel/addons/text/FlxTypeText.hx
+++ b/flixel/addons/text/FlxTypeText.hx
@@ -307,6 +307,18 @@ class FlxTypeText extends FlxText
 		_timer = 0;
 		_typing = false;
 		
+		if (useDefaultSound)
+		{
+			_sound.stop();
+		}
+		else
+		{
+			for (sound in sounds)
+			{
+				sound.stop();
+			}
+		}
+		
 		if (completeCallback != null)
 		{
 			completeCallback();


Gama11:
>Had this issue in one of my prototypes.

What issue exactly?

winterismute:
Basically, most of the typing sounds I was using were quite long, and typing might finish with the sound still playing for some noticeable amount of time. Not a big deal, it can be fixed by reworking the assets, but at the same time, I think it can be also logically ok to ensure in the code then when typing is finished no typing sound should be playing...

Gama11:
Can you share one as an example for testing?

winterismute:
Hi, I unfortunately can not share the projects with the issue (they are cool but they are basically gifts to my partner hence they are quite, well, personal). Making a new example would cost me, well, some time, and I am a bit short on that at the moment. However, every project with a typing text, release built, with a speed not too low and a sound not too short would generate the "problem" (=> hearing a sound for a short time after the text has finished).
Honestly, the issue is more visible when combined with this feature: https://github.com/HaxeFlixel/flixel-addons/pull/296
in one game, I use a fairly long sound of a typewriter writing (maybe 6 seconds) with flxtypetext. Cutting the single stroke sounds would be not only cumbersome, not only problematic since at low bitrate short mp3 sounds are badly encoded, but also, being them played randomly, I would not always have a credible typing sound below my text (tried it). Having a 6 seconds sound (that I do not reset thanks to the flag above), I need a mechanism to stop it when the text stops. But again, this is only an extreme example, not the only one.
This being said, I think there is also a strong logical point behind stopping the sound when the text completes inside the class, since the class is handling the sounds internally in all the other circumstances.

Gama11:
Nvm, I found a sound effect to test with: https://www.freesound.org/people/Natty23/sounds/257275/

PR #296, opened 2017-05-12
FlxTypeText: add finishSounds

winterismute:
Added flag "restartTypingSoundOnPlay" to control whether typing sounds should be stopped bofore playing again or not. Default is "true" which was the previous behavior.


Gama11:
Agreed, this makes a lot more sense for longer typing sounds. Not a fan of the naming though, I think I'll rename it to `completeSounds`.

Gama11:
Perhaps `finishSounds` is less ambigious, since there's a `completeCallback` etc already.

PR #297, opened 2017-06-04
add FlxTwinkleEffect. Handy for showing invincible state.

intersrc:


starry-abyss:
What is the difference with http://api.haxeflixel.com/flixel/effects/FlxFlicker.html ?

SeiferTim:
@starry-abyss it looks like this actually flashes the sprite White, instead of transparent (if I'm reading it right)

JoeCreates:
From the looks of the code I think this won't work on native due to the use
of the mask on copypixels?

intersrc:
@JoeCreates You are right. I'll try to solve it.

intersrc:
@JoeCreates Not perfect, but enough for pixel games.

intersrc:
@starry-abyss As @SeiferTim said, it can flash the sprite a specific color. And the color can be translucent, like `0x7fff0000`

Gama11:
There's a few code style issues here, and `#if cpp` is incorrect. It would have to be `#if sys` to also work on Neko. Not sure about HTML5 with WebGL rendering.

Besides, is that workaround even necessary anymore with OpenFL 8?

PR #312, opened 2018-04-13
Fix external tilesets for tiled

basxto:
Crash with a more meaningful error message when a path is wrong.
Support paths like `assets/data/maps/../terrains/dirt.tsx`.

### DIFF
--- a/flixel/addons/editors/tiled/TiledTileSet.hx
+++ b/flixel/addons/editors/tiled/TiledTileSet.hx
@@ -5,6 +5,7 @@ import flixel.util.typeLimit.OneOfTwo;
 import openfl.Assets;
 import openfl.utils.ByteArray;
 import haxe.xml.Fast;
+import haxe.io.Path;
 
 /**
  * Copyright (c) 2013 by Samuel Batista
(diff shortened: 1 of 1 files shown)


Gama11:
So Tiled generates paths that include `..`? That's a bit annoying.

flixel/addons/editors/tiled/TiledTileSet.hx
```diff
@@ -58,12 +59,16 @@ class TiledTileSet
...
-			var sourcePath = rootPath + source.att.source;
+			var sourcePath = haxe.io.Path.normalize(rootPath + source.att.source);
 			if (Assets.exists(sourcePath))
 			{
 				source = new Fast(Xml.parse(Assets.getText(sourcePath)));
 				source = source.node.tileset;
 			}
+			else
+			{
+				throw "Invalid TSX tileset path";
```
Gama11: This error should probably contain the `sourcePath`.

flixel/addons/editors/tiled/TiledTileSet.hx
```diff
@@ -58,12 +59,16 @@ class TiledTileSet
 		
 		if (source.has.source)
 		{
-			var sourcePath = rootPath + source.att.source;
+			var sourcePath = haxe.io.Path.normalize(rootPath + source.att.source);
```
Gama11: No need to fully qualify the path with `haxe.io` if you already imported the class. ;)

PR #314, opened 2018-05-01
Tiled: Guess rootpath from asset path

TomBebby:
A few additions to the TiledMap constructor so it guesses the root path

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   * fixed loading of external tile sets (#312)
   * improved error messages for invalid paths (#312)
   * fixed compatibility with Haxe 4
+* `FlxSlider`: fixed bounds not being updated after position changes (#306)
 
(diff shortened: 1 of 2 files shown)


flixel/addons/editors/tiled/TiledMap.hx
```diff
@@ -114,6 +120,13 @@ class TiledMap
 	{
 		for (node in source.nodes.tileset)
 		{
+			if (node.has.source)
+			{
+				var path = '${rootPath}/${node.att.source}';
+				path = REDUNDANT_PATH.replace(path, "/");
```
Gama11: Is `REDUNDANT_PATH` trying to do the same as `Path.normalize()` (which is used in https://github.com/HaxeFlixel/flixel-addons/pull/312)?

TomBebby:
Just updated to just the root path guessing - I was on the release version when I made the original changes so I missed that change

PR #315, opened 2018-05-01
TiledMap: guess rootPath based on data asset path

TomBebby:
I messed up my git repo so had to re-do this pull request

### DIFF
--- a/flixel/addons/editors/tiled/TiledMap.hx
+++ b/flixel/addons/editors/tiled/TiledMap.hx
@@ -46,7 +46,7 @@ class TiledMap
 	var noLoadHash:Map<String, Bool> = new Map<String, Bool>();
 	var layerMap:Map<String, TiledLayer> = new Map<String, TiledLayer>();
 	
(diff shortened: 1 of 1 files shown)


flixel/addons/editors/tiled/TiledMap.hx
```diff
@@ -45,24 +46,28 @@ class TiledMap
...
 		
 		if (Std.is(data, String)) 
 		{
+			if (this.rootPath == null)
+				this.rootPath = Path.directory(data) + "/";
 			source = new Fast(Xml.parse(Assets.getText(data)));
 		}
 		else if (Std.is(data, Xml)) 
 		{
+			this.rootPath = "";
```
Gama11: Hm.. is this correct? Maybe it should only be set if the `rootPath` parameter is `null`?

Gama11: I'll interpret the lack of a response as agreement. ;)

PR #318, opened 2018-05-22
FlxSliceSprite: add scale support, closes #317

### LINKED ISSUE #317: How should FlxSliceSprite handle scale?
Currently FlxSliceSprite is not affected by the state of its 'scale' property.

In my own build, I have its scale setting propagating to its internal renderSprite's scale variable, meaning that it stretches the final sprite, borders and all.

Should this be the correct implementation, or should it instead be taken into account when calculating width and height prior to regenerating its graphics, leaving the borders unstretched?

The way I see it, width etc should cause the slicing to regenerate, while scale should stretch it, but I won't PR that unless that's the desired use case.

Skerper:
Short of rewriting this extensively to use a callback point instead of a FlxPoint, this is the simplest way to make FlxSliceSprite implement scale as discussed in [#317](https://github.com/HaxeFlixel/flixel-addons/issues/317).

### DIFF
--- a/flixel/addons/display/FlxSliceSprite.hx
+++ b/flixel/addons/display/FlxSliceSprite.hx
@@ -357,6 +357,7 @@ class FlxSliceSprite extends FlxStrip
 		{
 			renderSprite.x = x;
 			renderSprite.y = y;
+			renderSprite.scale.copyFrom(scale);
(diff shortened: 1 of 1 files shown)


Gama11:
Thanks!

PR #320, opened 2018-06-15
TiledTileSet: add tile `type` and `probability` attributes

guifes:
Just adding these two neglected attributes present in tilesets.


Gama11:
Thanks!

PR #326, opened 2018-11-26
TiledTileSet: fix numRows, numCols

Birdwards:
These two variables were named backwards. Dividing the image height by the tile height yields the number of rows, not columns, and dividing the image width by the tile width yields the number of columns, not rows. I know this is a breaking change, but I guess it has to be if we want the names to be correct.


Gama11:
I'd say that's more of a bugfix than a breaking change. :)

PR #329, opened 2019-01-06
TiledImageLayer: document that x/y are for old Tiled versions

elliocheerio:
Removes old offset properties `x` and `y` from `TiledImageLayer`. As of Tiled version 0.15 the offsets are `offsetX` and `offsetY`, which are accessible from the parent class `TiledLayer`.

TLDR: [https://github.com/elliocheerio/haxeflixel-tiled-image-layers-fix-demo](https://github.com/elliocheerio/haxeflixel-tiled-image-layers-fix-demo)

### DIFF
--- a/flixel/addons/editors/tiled/TiledImageLayer.hx
+++ b/flixel/addons/editors/tiled/TiledImageLayer.hx
@@ -9,11 +9,18 @@ import haxe.xml.Fast as Access;
 class TiledImageLayer extends TiledLayer
 {
 	public var imagePath:String;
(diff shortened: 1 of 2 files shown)


Gama11:
Hm, perhaps it would be better to keep them for backwards-compatibility, but add doc comments to inform the user since which version they're no longer available? For instance, the TiledEditor demo no longer builds with this change (see Travis).

elliocheerio:
Woops! Keeping them for backwards-compatibility was my initial thought, should have gone with my gut ;) I've put comments in for which versions of Tiled they can be used for.

PR #333, opened 2019-07-01
Bring TiledObject in line with Tiled specs

JoeCreates:
This PR tweaks TiledObject to be more in line with Tiled maps.

Firstly it allows for multi line property values which are added as a string inside the element rather than as a `value` attribute.

Secondly it allows for float object positions, which is important for any object which is rotated in tiled to a non-90-degree multiple.

Finally it includes the tiled object's unique id (which is different to the gid which applies specifically to tiles).


Gama11:
Looks like this breaks compilation for several demos.

Gama11:
Still waiting on a flixel-demos PR.

JoeCreates:
I'm making some other little tweaks to the tiled stuff. Will ipdate when
I'm done with my current project.

Gama11:
Closing for now, please reopen when you get a chance to finish it with the corresponding fix to flixel-demos.

---

Btw, changing x and y from `Int` to `Float` is surprising to me, since the [docs for `<object>`](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#object) say:

>x: The x coordinate of the object **in pixels**.
>y The y coordinate of the object **in pixels**.

And if subpixels _are_ supported, wouldn't `width` and `height` have to be floats as well?

Issue #275, opened 2016-10-12
FlxSliceSprite slicing issues


Eiyeron:
Hi. Trying to use this class, I got to see a curious issue while using the last dev of flixel-addons (to have the sprite fix), there is an issue with the rightmost and bottom borders: they do not appear at all

``` haxe
//...
        background = new FlxSliceSprite(AssetPaths.textbox_slice__png, new FlxRect(6,8,32,32), 320, 16+NUM_LINES*21); // NUM_LINES = 3
        add(background);
//...
```

Here's an example of what it looks like (the right window is a neko build, which works). I don't have a clue on why it doesn't work, sadly.

![2016-10-12-234656_1338x546_scrot](https://cloud.githubusercontent.com/assets/936086/19328996/d3a32082-90d6-11e6-9cf3-7e75c1829a50.png)

Here's the used 9slice sprite
![textbox_slice](https://cloud.githubusercontent.com/assets/936086/19329082/436f8ba8-90d7-11e6-9f2c-e9951ad8bda6.png)

Have a nice day and thanks for your hard work!

Edit : After trying on Flash, I also got to see that using the exact same parameters, I get also an erroneous result on this target on the right side : two extra pixel columns are shown.
![screenshot_2016-10-14_14-29-30](https://cloud.githubusercontent.com/assets/936086/19387495/dcee9b72-921a-11e6-9beb-b63293dcfbd7.png)

Geokureli:
I'm having issues similar with FlxSliceSprite when playing on Safari.
<img width="620" alt="Screen Shot 2019-07-01 at 4 46 51 PM" src="https://user-images.githubusercontent.com/2609513/60469203-86be8f00-9c21-11e9-8aa8-8eb87057483e.png">
Compare the above with the result on Chrome
<img width="623" alt="Screen Shot 2019-07-01 at 4 57 50 PM" src="https://user-images.githubusercontent.com/2609513/60469237-a2c23080-9c21-11e9-91d4-76b9e1fa46a2.png">
Note that in this example the height of the sprite matches the source height, while the width is 1 greater

Beeblerox:
@Geokureli can you provide a texture you're using for slicing and having problem with?

Geokureli:
sure can
[...]

PR #334, opened 2019-07-13
Add sourceRect and fillCenter properties

Beeblerox:
`sourceRect` will allow users to define which part of image to use as a source for slicing.
`fillCenter` might help to save performance for cases when you don't need to fill center part of sliced sprite (leave it empty)


Gama11:
There's a lot of unrelated whitespace changes in the diff - we use [haxe-formatter](https://github.com/HaxeCheckstyle/haxe-formatter) now.

Beeblerox:
i'm sorry about that. didn't noticed them.
Hope that my second commit fixed it

Gama11:
Does this close #275?

Beeblerox:
it should, but i'd wait response from @Geokureli

PR #335, opened 2019-07-29
redo implementation of FlxSliceSprite

Beeblerox:
This will make it work correctly on old machines and in Safari browser.
Plus it doesn't generate new bitmapdatas for slice rects, so it could mean a bit less memory footprint

### DIFF
--- a/flixel/addons/display/FlxSliceSprite.hx
+++ b/flixel/addons/display/FlxSliceSprite.hx
@@ -224,6 +224,10 @@ class FlxSliceSprite extends FlxStrip
 		var numHorizontal:Int = Math.ceil((_snappedWidth - topLeft.width - topRight.width) / topMiddle.width);
 		var numVertical:Int = Math.ceil((_snappedHeight - topLeft.height - bottomLeft.height) / middleLeft.height);
 	
+		vertices = new DrawData<Float>();
+		uvtData = new DrawData<Float>();
+		indices = new DrawData<Int>();
+		colors = new DrawData<Int>();
 		numVertices = 0;
 
 		if (fillCenter)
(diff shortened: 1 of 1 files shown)


Geokureli:
Screen Shot 2019-08-05 at 4 35 25 PM
Noticing some issues on this branch where the right edge appears far off to the right.
[here's a video of it happening](https://imgur.com/v7Qo4Yb)

Beeblerox:
@Geokureli what are the params (width and height) of the object with this issue?

Geokureli:
I've gone and made [a testbed with the code that is causing the error](https://gist.github.com/Geokureli/cf3168efc9835208ae10b1e459f7c284). this should work with a default flixel project.xml
```xml
<window width="640" height="640" fps="60" background="#144491" hardware="true" vsync="true" />
```
these are the images (assets/images/ui/UiPellet.png & assets/images/ui/UiPellet_filled.png)
UiPellet UiPellet_filled

it seems to happen with a variety of sizes, but it didn't ever recreate the issue with with the original set of pellets. It's usually the second one, which makes me think the squish animation is what's causing this issue

Geokureli:
here's a smaller testbed that reproduces the issue:
https://gist.github.com/Geokureli/a512eca5a905626a5e0a5f8a92adc695

looks like if the sliceSprite's width shrinks by more than the right portion's width this issue happens

Beeblerox:
@Geokureli it could not hanle width/height less than SourceRect's width/height

Geokureli:
@Beeblerox it's never less than the source, in either of my examples

Beeblerox:
@Geokureli ok. i'll check it. (guess there is an error somewhere)

Beeblerox:
@Geokureli i've made small change which should help (i hope), please try it

Geokureli:
this works, thanks a lot!

Beeblerox:
@Geokureli no, thank you for your patience and code example!

PR #339, opened 2020-03-28
FlxTilemapExt: add setDownwardsGlue()

chosencharacters:
From Early Melon, who is answering my bounty.

This fix adds 10 lines and requires maxVelocity.y to be set and `setGlue(true)` on your tilemap

Example of this working can be found here:
https://github.com/chosencharacters/SlopeLand/tree/down-slope-fix

PS: Works best if the jump in your game has coyote time. When I put this into Renaine which has a pretty complicated move code, it didn't play friendly with velocities that are TOO high (as in, Renaine speed of something like 125) as they'll slide off the top, the fix for that is to ground your char as soon as they are within the tile for the first time then it works.

### DIFF
--- a/flixel/addons/tile/FlxTilemapExt.hx
+++ b/flixel/addons/tile/FlxTilemapExt.hx
@@ -410,7 +410,9 @@ class FlxTilemapExt extends FlxTilemap
 	 *
 	 * @param 	downwardsGlue  Activate/Deactivate glue on slopes on the
 	 * @param 	slopeSlowDownFactor  A slowing down factor while climbing slopes, from 0.0 to 1.0, By default 0.0, no slow down.
-	 * @param 	velocityYDownSlope The maximum velocity Y down a slope, it should be high enough to be able to use downwardsGlue. Default to 200
+	 * @param 	velocityYDownSlope The maximum velocity Y down a slope, it should be high enough to be able to use downwardsGlue. Default to 200.
+	 *
+	 * @since 2.9.0
 	 */
 	public function setDownwardsGlue(downwardsGlue:Bool, slopeSlowDownFactor:Float = 0.0, velocityYDownSlope:Float = 200):Void
 	{
--- a/flixel/addons/tile/FlxTilemapExt.hx
+++ b/flixel/addons/tile/FlxTilemapExt.hx
@@ -411,7 +411,7 @@ class FlxTilemapExt extends FlxTilemap
 	 * @param 	slopeSlowDownFactor  A slowing down factor while climbing slopes, from 0.0 to 1.0, By default 0.0, no slow down.
 	 * @param 	velocityYDownSlope The maximum velocity Y down a slope, it should be high enough to be able to use downwardsGlue. Default to 200
 	 */
-	public function setDownwardsGlue(downwardsGlue:Bool, slopeSlowDownFactor:Float, velocityYDownSlope:Float = 200):Void
+	public function setDownwardsGlue(downwardsGlue:Bool, slopeSlowDownFactor:Float = 0.0, velocityYDownSlope:Float = 200):Void
 	{
 		_downwardsGlue = downwardsGlue;
 		_slopeSlowDownFactor = 1 - slopeSlowDownFactor/10;
--- a/flixel/addons/tile/FlxTilemapExt.hx
+++ b/flixel/addons/tile/FlxTilemapExt.hx
@@ -39,7 +39,9 @@ class FlxTilemapExt extends FlxTilemap
 	var _slopePoint:FlxPoint = FlxPoint.get();
 	var _objPoint:FlxPoint = FlxPoint.get();
 	var _glue:Bool = false;
-
+	var _slopeSlowDownFactor:Float = 0;
+	var _velocityYDownSlope:Float = 0;
+	
 	var _slopeNorthwest:Array<Int> = [];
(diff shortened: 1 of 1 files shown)


Gama11:
What's with the commented out code?

Can the effects of this be seen in the FlxTilemapExt demo as well? If not, could we modify it a bit to show it?

jinjinjinK:
It works with FlxTilemapExt too, the commented code was a test to have different slow down while climbing according to the slope steepness, I got rid of it to have a universal slow down so adding the slope style is not necessary.

Gama11:
Commented code should be removed.

I just saw that this involves adding hardcoded keyboard input checks to `FlxTilemapExt`. That's not an acceptable fix / doesn't belong there.

chosencharacters:
Hey @Gama11 the key dependency has been removed. You can see this in action here:

https://github.com/HaxeFlixel/flixel-demos/pull/274

jinjinjinK:
The setGlue method has two parameters (with infos for users), setting glue to true or false to activate it  and a slow down factor (0.0 to 1.0) for climbing slopes.

chosencharacters:
I added a third so that you have the option of doing maxVelocity.y or some custom value

flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -735,6 +778,7 @@ class FlxTilemapExt extends FlxTilemap
 	function solveCollisionSlopeSoutheast(Slope:FlxObject, Object:FlxObject):Void
 	{
 		// Calculate the corner point of the object
+		if(Object.x<Slope.x-16) return;
```
Gama11: A hardcoded value of `16`? I'm guessing this only works with tilemaps that happen to use that as their tile size? ;)

Gama11:
Can you clarify a bit what exactly the intention of the changes is - at first I thought you had this `setGlue()` method and `if (_glue)` guards everywhere to make the new behavior completely opt-in and avoid breaking any existing behavior. But I just saw that a few of the added checks don't check for `_glue`, is that intentional? Are those lines different fixes unrelated to the "glue feature"?

Gama11:
Also, reading through https://github.com/chosencharacters/squidBounties/issues/2, it sounds like there already is "glue" in the current version, it's just not working as well as one would like. So a method named `setGlue()` might be a bit confusing, considering there's glue whether you have called it or not? :)

chosencharacters:
I'm confused. There is nothing in the current version of FlxTilemapExt that handles sticking to downwards slopes. What are you referring to?

Gama11:
>FlxTilemapExt has upward slopes and downward slopes. When going up slopes you are **glued** to the slope so you can smoothly and happily go upwards except for certain speeds where you hit "snags"

So glue isn't new, but glue for _downwards_ slopes is? Nothing about the name `setGlue()` indicates that, maybe it should be `setDownwardsGlue()`?

chosencharacters:
Oh, yea, that is a good point. We should do that.

jinjinjinK:
There is 2 commits, the glue method and var were added there by me.
There was no glue before that.
The hardcoded 16 is something I forgot. Bot lines in solveCollisionSlopeSouthwest & solveCollisionSlopeSoutheast should be removed.
I don't touch ceiling slopes.
Everything else that change the physics is in if(_glue) conditions.

jinjinjinK:
yes setDownwardsGlue makes more sense !

Gama11:
I'm not sure if it's related to those lines @jinjinjinK should be removed, but this seems to break collisions with ceiling slopes. This is what that looks like with the dev branch (working):



With this branch (broken - notice how you can get inside of the tiles):



Also, not sure what's going on at the end of that second .gif. :D

Gama11:
Ah, the second issue seems to require commenting out `setGlue()`.

chosencharacters:
Picking this up after an eternity, entirely my fault as Melon wrote these changes last month but I fell off the planet with it, but @Gama11 check this out

flixel/addons/tile/FlxTilemapExt.hx
```diff
@@ -391,14 +398,25 @@ class FlxTilemapExt extends FlxTilemap
...
 		}
 
 		return results;
 	}
-
+	/**
+	 * Set glue to force contact with slopes and a slow down factor while climbing 
+	 *
+	 * @param 	downwardsGlue  Activate/Deactivate glue on slopes on the
+	 * @param 	slopeSlowDownFactor  A slowing down factor while climbing slopes, from 0.0 to 1.0, By default 0.0, no slow down.
```
Gama11, line 411: Did you remove the default value for this parameter intentionally? Either the `@param` docs have to be updated, or the default value of `0` should be restored.

Gama11:
The two issues from my .gifs appear to be fixed now.

There's still some added code that's not behind an `if (_downwardsGlue)` check at the start of `solveCollisionSlopeNorthwest()` and `solveCollisionSlopeNortheast()`. Similar lines in `solveCollisionSlopeSouthwest()` and `solveCollisionSlopeSoutheast()` have been removed in the latest commit though. Is that intentional, or should the two remaining ones be removed as well?

chosencharacters:
Fixed @Gama11

jinjinjinK:
About the if(_downwardsGlue) you mentionned, I think it's better to leave those lines for all cases because it also fixed the issue for non Glue situation at slope's edge.

PR #341, opened 2020-04-17
FlxOgmo3Loader: add an optional tilemap parameter to loadTilemap()

01010111:


flixel/addons/editors/ogmo/FlxOgmo3Loader.hx
```diff
@@ -48,9 +48,9 @@ class FlxOgmo3Loader
 	 * @param	tileLayer		The name of the layer the tilemap data is stored in Ogmo editor, usually `"tiles"` or `"stage"`.
```
Gama11, line 48: Should probably update the docs.

Gama11:
Did the same in `FlxOgmo3Loader` as well for connsistency.

PR #342, opened 2020-05-22
Add FlxShaderMaskCamera

fratturobot:
This camera class adds the functionality described in https://github.com/chosencharacters/squidBounties/issues/1 using OpenFL's native support for shaders and DisplayObject masking.

A basic demo can be seen [here](https://www.fratturo.net/shadermaskdemo/).

Limitations:
- Is only supported on OpenFL 8 or higher.  
- Doesn't support alpha blending due to a limitation of OpenFL masks (see https://github.com/openfl/openfl/issues/1086).
- Only works in Tile render mode.

### DIFF
--- a/flixel/addons/display/FlxShaderMaskCamera.hx
+++ b/flixel/addons/display/FlxShaderMaskCamera.hx
@@ -1,0 +1,197 @@
+package flixel.addons.display;
+
+#if (openfl >= "8.0.0")
+
+import flixel.FlxBasic;
+import flixel.FlxCamera;
+import flixel.FlxG;
+import flixel.graphics.tile.FlxDrawBaseItem;
+import flixel.group.FlxGroup;
+import flixel.util.FlxColor;
+import openfl.display.Shader;
+import openfl.display.Sprite;
+import openfl.filters.ShaderFilter;
+import openfl.geom.Rectangle;
+
+/**
+ * An extension of FLxCamera that supports applying a shader to arbitrary sections of the game world.
+ * Only available in blit render mode.
+ * Does NOT currently support alpha blending; the effect will be applied to all pixels where the mask is not fully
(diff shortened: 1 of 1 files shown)


flixel/addons/display/FlxShaderMaskCamera.hx
```diff
@@ -0,0 +1,288 @@
...
+		shaderFilter = null;
+
+		super.destroy();
+	}
+
+	// Apply position/size changes to our duplicate and mask canvases
+	override function updateInternalSpritePositions():Void
+	{
+		super.updateInternalSpritePositions();
+		for (iCanvas in [shaderCanvas, maskCanvas])
```
Gama11, line 202: What's an `iCanvas`, is that a new Apple product I missed? :D

fratturobot: heh, should I change the name?  I've got two canvases I want to apply the same position to, and `canvas` is already a class variable.

Aurel300: Using `canvas` would just shadow the class variable inside the scope of the `for` loop.

fratturobot: Right, but isn't it better to avoid shadowing variables?  Since it can lead to confusion over whether a var is local or class-level if they share a name.

Aurel300: Well, as a general rule, it may be helpful, but here the loop is tiny (~6 lines), it is unlikely the reader would lose track of the iteration variable declared right there.

Gama11: The strange variable name definitely causes more confusion.

PR #343, opened 2020-06-13
FlxShaderMaskCamera: Revert to FlxCamera behavior in blit render mode

fratturobot:
A request from @chosencharacters , who wanted to be able to debug on the Flash target when using this class.  Now rather than throwing an exception, `FlxShaderMaskCamera` will simply act like a standard `FlxCamera` when `FlxG.RenderBlit` is true.


Gama11:
Why not just use a standard FlxCamera on Flash then?

chosencharacters:
I think we should avoid forcing workarounds and crashes as much as possible in flixel-addons, so doing this makes it a lot easier to use.

Gama11:
A crash could actually be preferable though because that informs you this functionality isn't available. This way it would just "fail" silently, right?

PR #348, opened 2021-01-16
TransitionFade: allow for the region attribute to work properly

ninjamuffin99:
setting the 'region' didn't work. Quick little fixins so now it does.

### DIFF
--- a/flixel/addons/transition/TransitionFade.hx
+++ b/flixel/addons/transition/TransitionFade.hx
@@ -7,6 +7,7 @@ import flixel.FlxSprite;
 import flixel.tweens.FlxTween;
 import flixel.util.FlxColor;
 import flixel.util.FlxGradient;
+import flixel.math.FlxRect;
 import openfl.Assets;
 import openfl.display.BitmapDataChannel;
 import openfl.geom.Matrix;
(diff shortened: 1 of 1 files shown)


ninjamuffin99:
just for example


even when setting the region data attribute, the size of the fade would always be hardcoded to FlxG.width and FlxG.height
image

now it fills whatever region you set it as
image

(camera is zoomed out in these screenshots, so if it's hardcoded to FlxG.width, it will have gaps, shown in the first image)

Gama11:
@larsiusprime Thoughts?

larsiusprime:
Seems reasonable

flixel/addons/transition/TransitionFade.hx
```diff
@@ -146,48 +146,48 @@ class TransitionFade extends TransitionEffect
 		}
 	}
 
-	function makeSprite(DirX:Float, DirY:Float):FlxSprite
+	function makeSprite(DirX:Float, DirY:Float, region:FlxRect):FlxSprite
```
Gama11, line 149: Um, CI seems to be failing because FlxRect isn't imported?

ninjamuffin99: woops! I just copy pasted my code from VS Code into Github web browser editor, forgot that!

notdonxd:
So this is accepted and fixed?

flixel/addons/transition/TransitionFade.hx
brosomethingwrongwithyothing: ok

PR #352, opened 2021-10-08
Allow finishCallback updates within callback

DigiEggz:
When changing finishCallback within an initial callback function, the new callback is set to `null` and erased on execution. This prevents that scenario by checking if the current callback is the same as the previously executed callback.

### DIFF
--- a/flixel/addons/transition/TransitionEffect.hx
+++ b/flixel/addons/transition/TransitionEffect.hx
@@ -62,13 +62,9 @@ class TransitionEffect extends FlxSpriteGroup
 		finished = true;
 		if (finishCallback != null)
 		{
-			var currentCallback = finishCallback;
-			finishCallback();
-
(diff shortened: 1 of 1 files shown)


Gama11:
What's the use case for that, does a `TransitionEffect` finish multiple times?

Metlmeta:
> 
> 
> What's the use case for that, does a `TransitionEffect` finish multiple times?

If you try to chain another finishCallback after its first callback, it won't execute because it's been nulled. As a pseudo-example:

```
transition.finishCallback = firstFunction;

function firstFunction()
{
    // ...
    
    // This callback will not be executed
    transition.finishCallback = secondFunction;
}

function secondFunction()
{
    // ...
}


```

Metlmeta:
Ahh, I misunderstood. I tested your suggestion and that was also working well.

PR #365, opened 2022-05-24
Fix FlxOgmo3Loader.getLevelValue

oscarcederberg:
According to the documentation, getLevelValue should read from the custom values/properties of the level-file. These are stored under the "Values"-objectin the JSON-file, not directly under root-object.


Geokureli:
re: #356
Great minds think alike
and we do as well

PR #368, opened 2022-07-18
Added FlxRuntimeShader

EliteMasterEric:
FlxRuntimeShader is a class which allows for compiling, running, and utilizing GLSL shaders at runtime, rather than at compile time.

The constructor of FlxRuntimeShader takes three arguments:
- `fragmentSource` is the text of the GLSL fragment shader as you would use. This is the same as what you would specify under `@:glFragmentSource`.
- `vertexSource` is the text of the GLSL vertex shader as you would use. This is the same as what you would specify under `@:glVertexSource`.
- `glslVersion` is the [GLSL version](https://en.wikipedia.org/wiki/OpenGL_Shading_Language#Versions) to use for the shader. Currently defaults to `120`, the minimum version needed to support uniforms.

Once constructed, you can assign the FlxRuntimeShader directly to the target sprite, like so:

```
var sprite:FlxSprite = new FlxSprite(0, 0);
var shader:FlxRuntimeShader = new FlxRuntimeShader(fragText);

sprite.shader = shader;

// The shader will now be called when rendering the sprite.
```

This class also supports retrieval and assignment of uniforms. To do this, use the provided functions:
- `getFloat(name:String):Null<Float>` returns the value of a `float` uniform if available, `null` otherwise.
- `setFloat(name:String, value:Float)` sets the value of a `float` uniform.
- `getFloatArray(name:String):Null<Float[]>` returns the value of a `float[]` uniform if available, `null` otherwise.
- `setFloatArray(name:String, value:Float[])` sets the value of a `float[]` uniform.
- `getInt(name:String):Null<Int>` returns the value of an `int` uniform if available, `null` otherwise.
- `setInt(name:String, value:Int)` sets the value of an `int` uniform.
- `getIntArray(name:String):Null<Int[]>` returns the value of an `int[]` uniform if available, `null` otherwise.
- `setIntArray(name:String, value:Int[])` sets the value of an `int[]` uniform.
- `getBool(name:String):Null<Bool>` returns the value of a `bool` uniform if available, `null` otherwise.
- `setBool(name:String, value:Bool)` sets the value of a `bool` uniform.
- `getBoolArray(name:String):Null<Bool[]>` returns the value of a `bool[]` uniform if available, `null` otherwise.
- `setBoolArray(name:String, value:Bool[])` sets the value of a `bool[]` uniform.

Vectors are treated as arrays of the corresponding length for the corresponding type. For example, to assign a `vec3` uniform, call `setFloatArray` and provide an array of three elements. Matrices are treated as arrays; to assign a `mat4` uniform, call `setFloatArray` and provide an array of 16 elements.

### DIFF
--- a/flixel/addons/display/FlxRuntimeShader.hx
+++ b/flixel/addons/display/FlxRuntimeShader.hx
@@ -1,0 +1,712 @@
+package flixel.addons.display;
+
+import flixel.system.FlxAssets.FlxShader;
+import lime.utils.Float32Array;
+import openfl.display.BitmapData;
+import openfl.display.ShaderInput;
+import openfl.display.ShaderParameter;
+import openfl.display.ShaderParameterType;
+
+/**
+ * An wrapper for Flixel/OpenFL's shaders, which takes fragment and vertex source
+ * in the constructor instead of using macros, so it can be provided data
+ * at runtime (for example, when using mods).
+ * 
+ * HOW TO USE:
+ * 1. Create an instance of this class, passing the text of the `.frag` and `.vert` files.
+ *    Note that you can set either of these to null (making them both null would make the shader do nothing???).
+ * 2. Use `flxSprite.shader = runtimeShader` to apply the shader to the sprite.
+ * 3. Use `runtimeShader.setFloat()`, `setBool()`, etc. to modify any uniforms.
+ * 
+ * @author MasterEric
+ * @see https://github.com/openfl/openfl/blob/develop/src/openfl/utils/_internal/ShaderMacro.hx
+ * @see https://dixonary.co.uk/blog/shadertoy
+ */
+class FlxRuntimeShader extends FlxShader
+{
+	#if FLX_DRAW_QUADS
+	// We need to add stuff from FlxGraphicsShader too!
+	#else
+	// Only stuff from openfl.display.Shader is needed
+	#end
+	// These variables got copied from openfl.display.GraphicsShader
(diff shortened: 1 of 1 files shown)


Geokureli:
I've been meaning to do something like this. I've wanted to make a shadertoy-esque shader editor that allows people to create and share custom flixel shaders. I think this is fantastic!

Geokureli:
I see flash is failing.

Also: is this something that will need to change periodically, perhaps due to changes to lime/openfl shaders? I see a lot of shared/similar code in FlxGraphicsShader, should we use shared code to keep these 2 classes in line with any new changes? I expecting, otherwise, that we will need to make the same changes in 2 places every time. 

Lastly, is there a unit test we could add that would let us know when thing DO need to be updated?

MasterEric:
I just pushed some new changes, which add `getBitmapData` and `setBitmapData` functions for setting `sampler2D` uniforms. This is useful for shaders which depend on multiple textures, such as shaders which apply certain blend modes.

I have also integrated the shader fixes which FlxRuntimeShader was applying into `openfl.display.Shader` directly. These changes eliminate any concerns with maintainability, but make this pull request dependent on the approval of openfl/openfl#2577.

The issues on the Flash platform have not yet been investigated nor diagnosed.

MasterEric:
Quick check-in on this PR, this code is tested and working, but builds will fail because it depends on openfl/openfl#2577, so this PR is pending the implementation of those changes.

PR #372, opened 2022-10-16
Fix the clipping issue in FlxBackdrop

NeeEoo:
This fixes a bug where when the camera is zoomed out, the backdrop stops rendering at the edges. All that needs to be done is `backdrop.camZoom = 0.5;` since making this run based on the current camera zoom would take up a lot of memory on blit

Fixes https://github.com/HaxeFlixel/flixel/issues/2649

### DIFF
--- a/flixel/addons/display/FlxBackdrop.hx
+++ b/flixel/addons/display/FlxBackdrop.hx
@@ -48,6 +48,19 @@ class FlxBackdrop extends FlxSprite
 	public var useScaleHack:Bool = true;
 
 	/**
+	 * The lowest zoom value that the backdrop will be able to have before it starts clipping.
+	 */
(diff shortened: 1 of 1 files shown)


Geokureli:
I still haven't seen a reliable repro for the issue you linked. Do you have simple way I can see it?

Also we shouldn't need a new property to fix 2649, blit targets are not a high priority, and {} brackets need to be on the following line

NeeEoo:
@Geokureli take the linked issue repro you were trying reproduce, then add this FlxG.camera.zoom = 0.5;

Geokureli:
@NeeEoo Thanks for explaining the cause of the issue, as for the issue, I went in a new direction and remade FlxBackdrop from the ground up in #373. It still needs some testing before merging, tho

PR #374, opened 2022-12-22
Fixed mismatched args in FlxClothSprite

UncertainProd:
In FlxClothSprite the arguments to `camera.drawTriangles` seem to be mismatched as there needs to be one more argument between 'blend' and 'antialiasing'. This should solve that.
camera.drawTriangles function signature: https://github.com/HaxeFlixel/flixel/blob/185c94aa73f7529d7411ab5cbf9f495297f94c75/flixel/FlxCamera.hx#L759

### DIFF
--- a/flixel/addons/effects/FlxClothSprite.hx
+++ b/flixel/addons/effects/FlxClothSprite.hx
@@ -200,7 +200,7 @@ class FlxClothSprite extends FlxSprite
 		if (_frameGraphic == null)
 			_frameGraphic = FlxGraphic.fromBitmapData(framePixels, false, null, false);
 
-		camera.drawTriangles(_frameGraphic, _vertices, _indices, _uvtData, colors, _point.addPoint(_drawOffset), blend, antialiasing);
+		camera.drawTriangles(_frameGraphic, _vertices, _indices, _uvtData, colors, _point.addPoint(_drawOffset), blend, false, antialiasing);
 	}
 
 	#if FLX_DEBUG


Geokureli:
Looking at the demo, everything seems fine here: https://haxeflixel.com/demos/FlxClothSprite/
Can you talk about the specific issue you're seeing and what this demo looks like with this change?

UncertainProd:
In the current state, setting antialiasing to either true or false had no effect on the sprite, but after this it should work
cloth-antialiasing

This was the flag png that I used:
flag

The cause for this was that the arguments were being passed incorrectly to camera.drawTriangles()
args

Code used:
```haxe
package;

import flixel.FlxG;
import flixel.FlxState;
import flixel.addons.effects.FlxClothSprite;

class PlayState extends FlxState
{
	var clothSprite:FlxClothSprite;

	override public function create()
	{
		super.create();
		clothSprite = new FlxClothSprite(FlxG.width / 2 - 105, 10, "assets/images/flag.png");
		clothSprite.pinnedSide = NONE;
		clothSprite.meshScale.set(1.5, 1);
		clothSprite.setMesh(15, 10, 0, 0, [0, 14]);
		clothSprite.iterations = 8;
		clothSprite.maxVelocity.set(200, 200);
		clothSprite.meshVelocity.y = 40;
		clothSprite.antialiasing = true;  // changing this before the fix didnt affect the sprite
		add(clothSprite);
	}

	override public function update(elapsed:Float)
	{
		super.update(elapsed);
		if (FlxG.mouse.pressed)
		{
			clothSprite.meshVelocity.x = FlxG.random.float(0, clothSprite.maxVelocity.x);
		}
		else
		{
			clothSprite.meshVelocity.x = 0;
		}
	}
}
```

Geokureli:
thanks!

PR #379, opened 2023-03-31
Add skipCallback to FlxTypeText

shimeji87:
Makes it easier to do actions if the text is skipped. Current work arounds is to have a boolean that is set when e.g the user presses any of the skipKeys or when the text is complete.


Geokureli:
I'm not following, skip is an immediate action, callbacks are for asynchronous actions, this can be achieved by simply calling a function after calling skip. 

For instance:

```hx
myTyper.skip(onTypeSkip);
```

would behave exactly the same as

```hx
myTyper.skip();
onTypeSkip();
```

PR #381, opened 2023-04-05
Rework on the FlxNestedSprite

Cheemsandfriends:
While I was using FlxNestedSprite for a project, I noticed that it was very... uncomfortable
for example, you would need to use a relative variable, which you would use in case you added a child, making it very difficult to do stuff like `sprite.setPosition()` or even `sprite.x`! but instead `sprite.relativeX`. same goes with acceleration, scale, etc.

This should fix the inconvenience.

### DIFF
--- a/flixel/addons/display/FlxNestedSprite.hx
+++ b/flixel/addons/display/FlxNestedSprite.hx
@@ -99,7 +99,7 @@ class FlxNestedSprite extends FlxSprite
 	var relativeColorTransform:ColorTransform;
 
 	/**
-	 * This will remove this sprite entirely. Use kill() if you
+	 * WARNING: This will remove this sprite entirely. Use kill() if you
 	 * want to disable it temporarily only and reset() it later to revive it.
 	 * Used to clean up memory.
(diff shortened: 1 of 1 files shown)


Cheemsandfriends:
> Can you type out examples of how things were done in the old code vs the new way?

Basically, when using FlxNestedSprite, if you added to a parent, the x value would be overriden by the parent position. same goes for alpha, angle, velocity, etc, etc...

With this, what it does is basically if the instance's a child, it would be concating the parent's "relativeMatrix" which would concat the grandparent's matrix and so forth until the beginning of its family tree lmao.

so in the old code, you would need to set `relativeX` to set the child's position, in the new one, you can do it directly with `x`. same goes for angle, every type of velocity and so on.

same goes with the color transform, so you can tint the sprite via `color`, put it inside a parent, and then it would be the tint + the parent's tint with `relativeColorTransform`.

itulau:
I've tested your changes and my sprite just went flying off the screen:

Screencast from 18-12-23 23 35 23

PR #383, opened 2023-04-11
FlxWeapon rework

DalekCraft2:
Clarified the outdated doc comments, removed unused fields, and changed how `fireRate` is implemented to make it more accurate to the game's time (because the old implementation didn't properly respect the game pausing when tabbed out). Also made `FlxWeapon` destroyable.

If this gets approved, I would also like to know whether I should destroy the `FlxSounds` or simply set them to `null`. Regardless of whichever one is picked, I'll still be making a commit to remove that TODO comment in the `destroy()` function.

### DIFF
--- a/flixel/addons/weapon/FlxWeapon.hx
+++ b/flixel/addons/weapon/FlxWeapon.hx
@@ -12,14 +12,14 @@ import flixel.math.FlxPoint;
 import flixel.math.FlxRandom;
 import flixel.math.FlxRect;
 import flixel.math.FlxVelocity;
+import flixel.tile.FlxTilemap;
+import flixel.util.helpers.FlxBounds;
+import flixel.util.helpers.FlxRange;
 #if (flixel >= "5.3.0")
 import flixel.sound.FlxSound;
 #else
 import flixel.system.FlxSound;
 #end
-import flixel.tile.FlxTilemap;
-import flixel.util.helpers.FlxBounds;
-import flixel.util.helpers.FlxRange;
 
 /**
  * A Weapon can only fire 1 type of bullet. But it can fire many of them at once (in different directions if needed) via createBulletPattern
@@ -171,10 +171,10 @@ class FlxTypedWeapon<TBullet:FlxBullet>
 	/**
 	 * Internal function that handles the actual firing of the bullets
 	 *
-	 * @param	Mode	Mode to use for firing the bullet
+	 * @param	mode	Mode to use for firing the bullet
 	 * @return	True if a bullet was fired or false if one wasn't available. The bullet last fired is stored in FlxWeapon.prevBullet
 	 */
-	function runFire(Mode:FlxWeaponFireMode):Bool
+	function runFire(mode:FlxWeaponFireMode):Bool
 	{
 		if (fireRate > 0 && FlxG.game.ticks < nextFire)
 		{
@@ -235,7 +235,7 @@ class FlxTypedWeapon<TBullet:FlxBullet>
 		currentBullet.elasticity = bulletElasticity;
 		currentBullet.lifespan = FlxG.random.float(bulletLifeSpan.min, bulletLifeSpan.max);
 
-		switch (Mode)
+		switch (mode)
 		{
 			case FIRE_AT_POSITION(x, y):
 				internalFireAtPoint(currentBullet, FlxPoint.weak(x, y));
@@ -331,12 +331,12 @@ class FlxTypedWeapon<TBullet:FlxBullet>
 	/**
 	 * Fires a bullet (if one is available) at the FlxTouch coordinates, using the speed set in setBulletSpeed and the rate set in setFireRate.
 	 *
-	 * @param	Touch	The FlxTouch object to fire at, if null use the first available one
+	 * @param	touch	The FlxTouch object to fire at, if null use the first available one
 	 * @return	True if a bullet was fired or false if one wasn't available. A reference to the bullet fired is stored in FlxWeapon.currentBullet.
 	 */
-	public function fireAtTouch(?Touch:FlxTouch):Bool
+	public function fireAtTouch(?touch:FlxTouch):Bool
 	{
-		var touch = Touch == null ? FlxG.touches.getFirst() : Touch;
+		var touch = touch == null ? FlxG.touches.getFirst() : touch;
 		if (touch != null)
 			return runFire(FIRE_AT_TOUCH(touch));
 		else
@@ -347,24 +347,24 @@ class FlxTypedWeapon<TBullet:FlxBullet>
 	/**
 	 * Fires a bullet (if one is available) at the given x/y coordinates, using the speed set in setBulletSpeed and the rate set in setFireRate.
 	 *
-	 * @param	X	The x coordinate (in game world pixels) to fire at
-	 * @param	Y	The y coordinate (in game world pixels) to fire at
+	 * @param	x	The x coordinate (in game world pixels) to fire at
+	 * @param	y	The y coordinate (in game world pixels) to fire at
 	 * @return	True if a bullet was fired or false if one wasn't available. A reference to the bullet fired is stored in FlxWeapon.currentBullet.
 	 */
-	public inline function fireAtPosition(X:Int, Y:Int):Bool
+	public inline function fireAtPosition(x:Int, y:Int):Bool
 	{
-		return runFire(FIRE_AT_POSITION(X, Y));
+		return runFire(FIRE_AT_POSITION(x, y));
 	}
 
 	/**
 	 * Fires a bullet (if one is available) at the given targets x/y coordinates, using the speed set in setBulletSpeed and the rate set in setFireRate.
 	 *
-	 * @param	Target	The FlxSprite you wish to fire the bullet at
+	 * @param	target	The FlxSprite you wish to fire the bullet at
 	 * @return	True if a bullet was fired or false if one wasn't available. A reference to the bullet fired is stored in FlxWeapon.currentBullet.
 	 */
-	public inline function fireAtTarget(Target:FlxSprite):Bool
+	public inline function fireAtTarget(target:FlxSprite):Bool
 	{
-		return runFire(FIRE_AT_TARGET(Target));
+		return runFire(FIRE_AT_TARGET(target));
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


Geokureli:
This has breaking changes on a niche, legacy class. At a glance, I see odd things I wouldn't recommend, like using a FlxTimer to control fire rate when I would use a float and add elapsed on update. I recommend making a new class in a new repo and battle-testing it in your game project over the course of it's development and then we can talk about replacing this entirely in a new major version.

FlxWeapon was ported from the flash version of flixel and it isn't used very often now. If I were making a shoot 'em up, I would probably just make my own system. Looking at the old version and you're version, they're pretty different from my version would look that, and that's fine, but it makes me thing we shouldn't even have an "official haxeflixel version" and just make little repos that people can copy

Geokureli:
Can you list the changes made, @DalekCraft2 ?

DalekCraft2:
You mean how some commits contain changes what were not mentioned in the commit's name?

DalekCraft2:
Here are the changes what I made and why I made them:

- Changed function arguments to start with lowercase letters because I'm pretty sure you're already trying to do that with all of Flixel
- Went through all of the doc comments and removed references to outdated functions because those won't be helping anyone
- Removed unused fields because they were likely replaced by `speedMode` and `fireFrom` at some point, though I could have also given them accessors what reference their respective fields in the enums
- Made `FlxWeapon` implement `IFlxDestroyable` because it has a few fields what can be `put()` or `destroy()`ed
- Changed the implementation of `fireRate` to use an `FlxTimer` because the current implementation did not respect how the game pauses when it is tabbed out
- Overrode `FlxBullet#destroy()` so it calls `bounds = FlxDestroyUtil.put(bounds)`, to be safe
- Changed some `FlxPoint`-related things in some functions like `runFire()` and `rotatePoints()` to ensure that no `FlxPoint` would be taken from the pool but not returned

flixel/addons/weapon/FlxWeapon.hx
```diff
@@ -285,16 +271,20 @@ class FlxTypedWeapon<TBullet:FlxBullet>
...
+	 * @param	returnedPoint An optional point to reuse instead of getting another from the pool.
+	 * @return	The new rotated point.
 	 */
-	public function rotatePoints(point:FlxPoint, origin:FlxPoint, angle:Float):FlxPoint
+	public function rotatePoints(point:FlxPoint, origin:FlxPoint, angle:Float, ?returnedPoint:FlxPoint):FlxPoint
 	{
-		var returnedPoint:FlxPoint = FlxPoint.weak();
+		if (returnedPoint == null)
+		{
+			returnedPoint = FlxPoint.weak();
```
Geokureli, line 286: functions shouldn't return a weak point, no idea why the old code did that
```suggestion
			returnedPoint = FlxPoint.get();
```

flixel/addons/weapon/FlxWeapon.hx
```diff
@@ -55,72 +56,58 @@ class FlxTypedWeapon<TBullet:FlxBullet>
...
-	 * Only accessible when fireFrom is "PARENT"
+	 * A timer used whenever the weapon is fired, with its duration set to `fireRate`.
 	 */
-	public var parent(default, null):FlxSprite;
+	public var fireTimer:FlxTimer;
 
 	/**
-	 * If true, when fired the bullet direction is based on parent sprites facing value (up/down/left/right)
-	 */
-	public var useParentDirection:Bool;
```
Geokureli, line 90: We can't just remove vars, we deprecate them first but leave them functioning until the next major version

flixel/addons/weapon/FlxWeapon.hx
```diff
@@ -86,22 +86,6 @@ class FlxTypedWeapon<TBullet:FlxBullet>
...
-	/**
-	 * If `parent == null`, the Weapon will fire from a fixed position on the screen, like in the game Missile Command.
-	 */
-	public var firePosition(default, null):FlxBounds<FlxPoint>;
-
-	/**
-	 * A value to use to offset a bullet's position when it is fired.
-	 * Can be used to, for example, line a bullet up with the "nose" of a space ship.
-	 */
-	public var positionOffset(default, null):FlxPoint;
```
Geokureli: same here, just deprecate them

flixel/addons/weapon/FlxWeapon.hx
```diff
@@ -55,72 +56,58 @@ class FlxTypedWeapon<TBullet:FlxBullet>
...
-	 * TODO - Needs testing with a scrolling map (when not using single screen display)
-	 */
-	public var bounds:FlxRect;
-
-	/**
-	 * Only accessible when fireFrom is "PARENT"
+	 * A timer used whenever the weapon is fired, with its duration set to `fireRate`.
 	 */
-	public var parent(default, null):FlxSprite;
+	public var fireTimer:FlxTimer;
```
Geokureli, line 76: I would just revert this aspect of the PR, I don't like using a FlxTimer for this, it should be a private float.

 I would just rewrite FlxWeapon to extend FlxTypedGroup and do it from the ground up in a public repo until it's solid enough to replace this in the next major version.
 
 Truthfully though, If FlxWeapon was written now I probably wouldn't put it in flixel-addons to begin with, it should just be it's own isolated demo that people can use if they like that implementation. If you do rewrite it in a new repo and it's decent I'd rather just remove this entirely and put yours in [External Tools](https://github.com/HaxeFlixel/flixel/wiki/External-Tools)

SeiferTim:
I'm going to side with @Geokureli here and say that `FlxWeapon` is kind of an outlier and should probably be removed altogether.
It sort of implies that using this class is 'the right way' to do projectiles in your game, and, in reality, there are many, many different ways that might fit and individual project better than trying to use this class - especially since it breaks so many of the normal conventions that are staples in HF.
If we could get away with removing it, I would love to have maybe a Sample that shows some of the more modern techniques to doing things like this for people to base their code on rather than have a class that is a 'cookie cutter' solution for everyone.
Maybe even a section in Snippets dedicated to 'Game Mechanics' and have some samples of weapons in there.

Doing a little poking around in `flixel-addons` there are a couple of other classes grandfathered over from AS3 Flixel and Photostorm that I think might not be working properly anymore (FlxStarfield seems to be not working?)

Geokureli (changes requested):
no breaking changes

flixel/addons/weapon/FlxWeapon.hx
DalekCraft2: I'm instead making them actually access the fields in the enums, but there is an issue: `positionOffset` is of type `FlxPoint` whereas `FlxWeaponFireFrom.POSITION.position` is of type `FlxBounds<FlxPoint>`. What should I do about this?
Also, do you know whether it's possible to change the fields of an enum without making another "instance" of one?

flixel/addons/weapon/FlxWeapon.hx
DalekCraft2: So do you want me to go back to the old implementation, or to try to do one of these other suggestions what you have listed?

Geokureli:
> I'm going to side with @Geokureli here and say that `FlxWeapon` is kind of an outlier and should probably be removed altogether. It sort of implies that using this class is 'the right way' to do projectiles in your game, and, in reality, there are many, many different ways that might fit and individual project better than trying to use this class - especially since it breaks so many of the normal conventions that are staples in HF. 

Agree with all of this.

> Doing a little poking around in `flixel-addons` there are a couple of other classes grandfathered over from AS3 Flixel and Photostorm that I think might not be working properly anymore (FlxStarfield seems to be not working?)

Got more examples?

DalekCraft2:
I did not reimplement one of the variables, `lastFired`, because it was private and not used anywhere, so let me know if I should add that back.

Geokureli:
@DalekCraft2 I made some changes

Passed in objects used as the return value are typically named `result`, I also simplified `rotatePoints`.

when editing doc comments use the following formatting:

- no tabs after the left "*" margin
- 3 spaces between `@param` and the param name
- with spaces, indent all param descriptions to be 2 space characters right of the longest param
- try to keep lines under ~120 characters, count tab indents as 4 characters (vscode should tell you the Col in the bottom right)
- After you've decided to split a long line try to keep the resulting lines under 100 columns (this does not mean that all lines must be under 100 columns!)
- when splitting param descriptions into multiple lines, indent the second line, with spaces so it starts at the same column as the first
- 2 spaces after `@return`
- no need to indent the second line of a return description, since there's only one return value
- I should add this to the formatting guide

here's an example, in my VSCode editor with "Render Whitespace" set to "boundary'

Screenshot 2023-04-13 at 4 34 52 PM

flixel/addons/weapon/FlxWeapon.hx
```diff
@@ -51,106 +51,126 @@ class FlxTypedWeapon<TBullet:FlxBullet>
...
+	 */
 	public var onPostFireSound:FlxSound;
 
 	/**
-	 * The factory function to create a bullet
+	 * The factory function used to create new bullets.
 	 */
 	var bulletFactory:FlxTypedWeapon<TBullet>->TBullet;
 
-	var lastFired:Int = 0;
```
Geokureli, line 136: removing this is technically a breaking change but i'll allow it. In all likelyhood no real person will have a compile error unless they use this field in an extending class, and highly unlikely and easy to fix

Geokureli (approved):

Geokureli:
when making these changes did you have a project you used to test them out? If so, is it public? Can you share it?

I try to make [a public test state](https://github.com/Geokureli/flixel-tests/tree/main/Source/states) for every feature, fix, or improvement I make to flixel, and use it to help others, or make unit tests. Here's an example: https://github.com/HaxeFlixel/flixel/pull/2768

If you don't have a small one, I'll make my own, but it might take a bit

DalekCraft2:
I had originally made the `FlxTimer`/`IFlxDestroyable` code in Quench and then decided to try to make a PR for them, so I knew that those worked already. I agree that it would be better to do a less complicated test, though.

Also, it would probably be a bad idea for *you* to test it with Quench because I use my awful Git habits over there constantly. By this point, every Haxe file in the VS Code directory viewer has a yellow name, meaning the version what you would see on the GitHub page would be quite outdated.

DalekCraft2:
So what is the consensus for this PR?

Geokureli:
I still need to test the changes and I haven't had time because FlxWeapon is pretty low on my list of priorities. My opinion of FlxWeapon is that it probably shouldn't exist and some new standalone tool should be made from the ground up, however I don't want to remove it until the next major version release, which isn't any time soon so I'm allowing this change if it doesn't have any problems.

Geokureli:
I finally tested this. Lot of issues. 

1. I get deprecation warnings if my project uses FlxWeapon, because FlxWeapon itself uses these classes.
2. I get runtime errors for what seems like a basic implementation
3. I find FlxWeapon to be incredibly confusing, it's possible I'm doing something wrong but it's so cumbersome to even set up a basic weapon that fires in the direction the parent is facing

Here's in my example, so far:
```hx
    var weapon:FlxWeapon;
    var gun:FlxSprite;
    
    override function create()
    {
        super.create();
        
        add(gun = new FlxSprite(100, 100, "assets/images/haxe.png"));
        weapon = new FlxWeapon
            ("test", (w)->new FlxBullet()
            , PARENT
                ( gun
                , new FlxBounds(FlxPoint.get(0, 0), FlxPoint.get(gun.frameWidth, gun.frameHeight))
                , true
                )
            , SPEED(new FlxBounds(5.0, 10.0))
            );
    }
```

I get the sense these changes weren't tested, the error happens because `set_fireFrom` sets `useParentDirection` which calls `set_useParentDirection` which crashes because it's checking fireFrom, which hasn't been set yet. I tried looking through quench, but there's nothing to grab thats easy to duplicate for testing.

Truthfully, after have a taste of FlxWeapon I leaning towards retiring this feature. I think it's poorly executed and never have been in flixel to begin with. I also hate FlxBounds and want to remove all references to that. I'm even considering removing the tool in the next major version, I would create a new repo called LegacyFlxWeapon that people can use if they have old projects that need it, either that or just deprecate the entire class or put a huge disclaimer on it to prevent people from creating issues or PRs to this. It's just too niche, and it's not worth my time compared to just about any other flixel feature

I think you can make a far better weapon/bullet manager than this, and you absolutely should, and we'll add it to the external tools list.

Geokureli:
If i were to try and save this class, I would do [something like this](https://github.com/HaxeFlixel/flixel-addons/compare/george-weapon?expand=1), but I still think the class is trash and should just be rewritten in a new repo

PR #385, opened 2023-04-14
`FlxControlHandler` TODO cleanup

DalekCraft2:
I cleaned up a few TODO comments in `FlxControlHandler` by doing what they requested. Some of them involved going back to the [original `FlxControlHandler.as`](https://github.com/photonstorm/Flixel-Power-Tools/blob/dev/Test%20Suite/src/org/flixel/plugin/photonstorm/FlxControlHandler.as) and adding the functions what were not present in the HaxeFlixel port.

Note: I had to sort of guess how `_xSpeedAdjust` and `_ySpeedAdjust` were going to be implemented in the original AS3 version because they were only set to `0` in `resetSpeeds()` and their values were never read. The arguments of `speedUp()` and `slowDown()` suggested they would be factors, but having them be `0` by default would make the speed constantly `0`, so I decided to add them to the speed instead.

I also added three more TODO comments regarding changes what I wanted to make but suspected would have made the Git diff much larger.

<sup>i'm on a roll here</sup>

### DIFF
--- a/flixel/addons/plugin/control/FlxControlHandler.hx
+++ b/flixel/addons/plugin/control/FlxControlHandler.hx
@@ -4,14 +4,12 @@ package flixel.addons.plugin.control;
 import flash.geom.Rectangle;
 import flixel.FlxG;
 import flixel.FlxSprite;
+import flixel.math.FlxAngle;
 import flixel.math.FlxMath;
 import flixel.math.FlxPoint;
 import flixel.math.FlxVelocity;
-#if (flixel >= "5.3.0")
 import flixel.sound.FlxSound;
-#else
-import flixel.system.FlxSound;
-#end
+import haxe.Constraints.Function;
 
 /**
  *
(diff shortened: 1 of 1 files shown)


Geokureli:
Honestly, this is another class that is a relic from flash flixel that I don't even recommend using and might remove in the next major version

Geokureli:
I've honestly never heard of anyone using this, and just recommend making your own version in a new repo, because aagain, this is a niche legacy feature that can be dozen a million difference ways and should have a 1-size-fits-all official implementation

Geokureli:
a lot of the input features you add should just use FlxActions instead

Geokureli:
I'll say it again. please just focus on making games in Haxeflixel. ask questions when you think some flixel tool is lacking and needs repair, ask for help in solving problems you're having while making your game, make github issues describing odd things you see that should be fixed. If a system isn't to your liking, start making your own tools in your own repo, with the focus of getting you're own game done, and after it has some actual use, maybe then we can talk about making changes to existing systems and adding new features.

I think you're still learning flixel, and really just want to leave your mark on it and it's gonna take time to learn how to make meaningful changes.

please just create issues about things you think should change before changing anything, and PLEASE focus on making games and learning flixel before changing it

PR #391, opened 2023-06-23
Don’t render text you can’t see

MemeHovy:
This should prevent some lag issues happening, this has happened in a FNF mod I was making where the text would stat to go off screen and the frame rate would start to drop, so this PR should help prevent this from happening.

(Currently a draft since I can’t test it yet.)

### DIFF
--- a/flixel/addons/text/FlxTypeText.hx
+++ b/flixel/addons/text/FlxTypeText.hx
@@ -547,4 +547,10 @@ class FlxTypeText extends FlxText
 		_sound.loadEmbedded(new TypeSound());
 		#end
 	}
+
+	override function regenGraphic(){
+		if (textField == null || !_regen || !isOnScreen())
+			return;
+		super.regenGraphic();
+	}
 }


SeiferTim:
This looks like it would not work if the Text is using a non-default camera.

Geokureli:
Tim is correct, if you want a textfield that doesn't redraw when not displayed than extend it with ad-hoc functionality like this.

Why do you want this? why change a text field that isn't added to a camera?

MemeHovy:
> Tim is correct, if you want a textfield that doesn't redraw when not displayed than extend it with ad-hoc functionality like this.
> 
> Why do you want this? why change a text field that isn't added to a camera? My guess is that you're using FlxTypeText to control a FlxBitmapText, if so, extend FlxTypeText and make it never redraw. or take the functionality out of FlxTypeText and put it in a third party class that manipulates either text class

Thanks for the suggestions, I wanted this since it was causing lag when a lot of text was displayed outside the screen, even though it should not draw it when it’s not seen.

Geokureli:
honestly this is another reason why we should improve FlxBitmapText.

Closing this, this isn't the right fix. make an issue for the problem you're having and we can discuss better options, there

MemeHovy:
> honestly this is another reason why we should improve FlxBitmapText.
> 
> Closing this, this isn't the right fix. make an issue for the problem you're having and we can discuss better options, there

Alright then

MemeHovy:
Made the issue here. #393

PR #405, opened 2023-09-02
Flx extended sprite

Geokureli:
This class sucks, it should be removed, but lets remove deprecation warnings for flixel 5.4.0


MemeHovy:
> This class sucks, it should be removed, but lets remove deprecation warnings for flixel 5.4.0

I mean, why not remove it now if it sucks?

Geokureli:
that's more work and might cause people to not update flixel if there are not viable replacements. also typically in open source software you don't just remove features unless it prevents new better features. I think we can make FlxMouseSpring work without needed FlxExtended sprite, then we can deprecate it, and then later it can be removed when people who used it have had time to change old projects

PR #412, opened 2023-12-14
Fix transition bug with persistent images

NeeEoo:
Also optimized the fade transition, by scaling the 0 angle version


Geokureli (changes requested):
can you test this with `FlxSprite.defaultAntialiasing = true`, if it looks bad then you will need to set `sprite.antialiasing = false;`

NeeEoo:
Should probably be better to be safe, with having it set antialiasing to false, because it might change on different platforms like between windows, mac, linux, html, flash, and i cant really test that

NeeEoo:
Not sure if i should add `sprite.antialiasing = true;` in the diagonal wipe, since that one has no 1px wide scaling

I also placed the no antialiasing to affect the other wipes, since they are also optimized with 1px and scaled

Geokureli (approved):

Geokureli:
Thanks!

PR #414, opened 2023-12-20
Fixed `FlxRuntimeShader` on the latest flixel commit.

MAJigsaw77:
On the latest commit in flixel repo, the references to `FLX_DRAW_QUADS` got removed in [this](https://github.com/HaxeFlixel/flixel/commit/5ebdcbb443ca2c0b095c62395cf312b0984269f6) commit and now the define is deprecated, so i made this pr to fix it

### DIFF
--- a/flixel/addons/display/FlxRuntimeShader.hx
+++ b/flixel/addons/display/FlxRuntimeShader.hx
@@ -1,10 +1,8 @@
 package flixel.addons.display;
 
-#if nme
-#error "FlxRuntimeShader is not available with nme"
-#elseif flash
-#error "FlxRuntimeShader is not available when targeting flash"
-#else
+#if (FLX_NO_COVERAGE_TEST && (nme || flash))
+#error "FlxRuntimeShader is not available with nme or flash"
+#end
 import flixel.graphics.tile.FlxGraphicsShader;
 #if lime
 import lime.utils.Float32Array;
@@ -652,4 +650,3 @@ class FlxRuntimeShader extends FlxGraphicsShader
 		return __glVertexSource = value;
 	}
 }
-#end
--- a/flixel/addons/display/FlxRuntimeShader.hx
+++ b/flixel/addons/display/FlxRuntimeShader.hx
@@ -1,7 +1,9 @@
 package flixel.addons.display;
 
-#if (FLX_NO_COVERAGE_TEST && (nme || flash))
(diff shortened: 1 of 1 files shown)


flixel/addons/display/FlxRuntimeShader.hx
```diff
@@ -1,6 +1,6 @@
 package flixel.addons.display;
 
-#if (FLX_DRAW_QUADS && !flash)
+#if (!nme && !flash)
```
Geokureli, line 3: this change makes sense but it was setup wrong before and we should fix it while we're here.

```suggestion
#if nme
    #error "FlxRuntimeShader is not available with nme"
#elseif flash
    #error "FlxRuntimeShader is not available when targeting flash"
#else
```

Geokureli: also note that the `#else` can just be an `#end` to simplify (it doesn't seem to compile the rest of the file once `#error` is reached

MAJigsaw77: I've been using these for all my native C++ libs, it's pretty useful but i never though about using it here.

Geokureli:
i see some errors, I might have been wrong about it needing to reference it. I think because coverage tests automatically try to reference everything, I'll see if theres a way around this

flixel/addons/display/FlxRuntimeShader.hx
```diff
@@ -1,6 +1,10 @@
 package flixel.addons.display;
 
-#if (FLX_DRAW_QUADS && !flash)
+#if nme
+#error "FlxRuntimeShader is not available with nme"
+#elseif flash
+#error "FlxRuntimeShader is not available when targeting flash"
+#else
```
Geokureli, line 7: we need to not throw errors when FLX_COVERAGE_TEST is defined, maybe

```suggestion
#if FLX_NO_COVERAGE_TEST
#if nme
#error "FlxRuntimeShader is not available with nme"
#elseif flash
#error "FlxRuntimeShader is not available when targeting flash"
#end
#end
```

Geokureli: or simplify with
```hx
#if (FLX_NO_COVERAGE_TEST && (nme || flash))
#error "FlxRuntimeShader is not available with nme or flash"
#end
```

Geokureli:
thanks!

Issue #413, opened 2023-12-19
FlxNestedSprite show flixel default icon when no image loaded


itulau:
I've noticed FlxNestedSprite shows the Flixel default icon when there isn't an image nor graphic loaded which seems to be the default behavior of `draw()` inherited from FlxSprite. So every FlxNestedSprite that itself doesn't render anything but contains children will have a default icon hidden within it by default. 
![Screenshot from 2023-12-19 16-30-16](https://github.com/HaxeFlixel/flixel-addons/assets/102986935/b263589f-6f6b-456b-902d-1ac8ea398ca9)

flixel-addons: 3.2.1
flixel:5.5.0

Geokureli:
is this happening on flixel 5.5.0 or are you using the dev github branch, or an older version of flixel?

itulau:
Edited my comment, but I'm on flixel 5.5.0

itulau:
Could reproduce in latest dev version of flixel/flixel-addons

Geokureli:
thanks for clarifying. Truthfully FlxNestedSprite is not actively maintained, or well-understood by the current team, so we tend to recommend FlxSpriteGroup as a substitute. In fact, it would be very helpful to hear why you chose it, and how you're using it. I've thought about deprecating this class, since it seems so niche and problematic

if you wanted to make a PR to fix this, I think it makes sense, and (in my limited knowledge of FlxNestedSprite) seems like an easy change

itulau:
Hi, I've noticed that FlxNestedSprite works better for the use-case I need right now: a group of sprites that represents a character where one sprite acts as the collider and the rest of the sprites are just visual and do not interact with the collisions. 
To my limited knowledge from my testings, it seems that FlxSpriteGroup was not designed for this because when collision occurs, the rest of the sprites simply ignore this. With FlxNestedSprite I could get the result I was looking for more or less with the same code.

FlxSpriteGroup:
 
![Peek 2023-12-20 15-34](https://github.com/HaxeFlixel/flixel-addons/assets/102986935/7f4f2351-5f95-4a51-81b7-eebf9bd9965c)

FlxNestedSprite:

![Peek 2023-12-20 15-36](https://github.com/HaxeFlixel/flixel-addons/assets/102986935/eced3ac6-e4ae-4847-b4b1-18a25133d79b)

[...]

PR #415, opened 2023-12-22
Fix FlxNestedSprite drawing default png by default

itulau:
FlxNestedSprite shows the Flixel default icon when there isn't an image nor graphic loaded which seems to be the default behavior of draw() inherited from FlxSprite. So every FlxNestedSprite that itself doesn't render anything but contains children will have a default icon shown by default.

This PR adresses this problem by guarding the call for drawing of the current FlxNestedSprite with a null check for _frame (the same variable that's used when checking for null when adding the default icon).

Solves #413


Geokureli:
Thanks!

PR #421, opened 2024-02-03
Add support for clipRects to FlxTiledSprite.

EliteMasterEric:
Adds the ability to clip `FlxTiledSprites` inside a rectangle by assigning the `clipRect`.

### DIFF
--- a/flixel/addons/display/FlxTiledSprite.hx
+++ b/flixel/addons/display/FlxTiledSprite.hx
@@ -8,6 +8,7 @@ import flixel.math.FlxMath;
 import flixel.system.FlxAssets.FlxGraphicAsset;
 import flixel.util.FlxColor;
 import flixel.util.FlxDestroyUtil;
+import flixel.math.FlxRect;
 import flixel.util.FlxSpriteUtil;
 
 /**
@@ -104,6 +105,14 @@ class FlxTiledSprite extends FlxStrip
 	{
 		graphic = FlxGraphic.fromFrame(Frame);
 		return this;
+	}
+
+	override function set_clipRect(Value:FlxRect):FlxRect
+	{
+		if (Value != clipRect)
+			regen = true;
+
+		return super.set_clipRect(Value);
 	}
 
 	override function set_graphic(Value:FlxGraphic):FlxGraphic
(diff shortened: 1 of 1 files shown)


Geokureli:
forgot this, i'll test this out tomorrow

flixel/addons/display/FlxTiledSprite.hx
```diff
@@ -106,6 +107,14 @@ class FlxTiledSprite extends FlxStrip
...
 	}
 
+	override function set_clipRect(Value:FlxRect):FlxRect
+	{
+		if (Value != clipRect)
+			regen = true;
+
+		return super.set_clipRect(Value);
+	}
+
```
Geokureli, line 117:
```suggestion
	override function set_clipRect(value:FlxRect):FlxRect
	{
		regen = true;

		return super.set_clipRect(value);
	}

```

couple things
1. use lower case arg names
2. regen even if the same rect instance is set. one common way to use this is

```hx
sprite.clipRect.width = 100;
sprite.clipRect = sprite.clipRect;
```

Geokureli (changes requested):
this doesn't seem to work right. In addition to the change I requested above, it behaves in unexpected ways here:

https://github.com/HaxeFlixel/flixel-addons/assets/2609513/036ed448-2a17-4d1e-af7a-bd5fba5e1959

notice that the bottom is cut off, and it tiles above when rect.y=-50 and height=200. I wasn't expecting it to tile above, but I was expecting it to look identical to the initial state in the vid

Here is my test case (it requires the above change to work):
```hx
package states;

import flixel.text.FlxText;
import flixel.math.FlxRect;
import flixel.addons.display.FlxTiledSprite;
import flixel.FlxG;

class TiledSpriteTestState extends flixel.FlxState
{
    var sprite:FlxTiledSprite;
    var text:FlxText;
    
    var toggled = true;
    
    override function create()
    {
        super.create();
        
        sprite = new FlxTiledSprite("assets/images/haxe.png", 200, 100);
        sprite.clipRect = FlxRect.get(25, 0, 150, 100);
        sprite.screenCenter();
        add(sprite);
        
        text = new FlxText("");
        updateText();
        add(text);
    }
    
    function updateText()
    {
        text.text = sprite.clipRect.toString();
        text.screenCenter();
        text.y += 100;
    }
    
    override function update(elapsed)
    {
        super.update(elapsed);
        
        if (FlxG.keys.justPressed.SPACE)
        {
            sprite.clipRect = toggled ? sprite.clipRect.set(0, 0, 200, 100) : sprite.clipRect.set(25,-50,150,200);
            toggled = !toggled;
            updateText();
        }
    }
}
```

PR #426, opened 2024-04-12
Avoid FlxBackdrop matrix leak

MaybeMaru:
Creates a FlxMatrix and FlxPoint variable at the start to prevent creating a FlxMatrix and getting a FlxPoint from the pool every frame on drawComplex.
Also makes sure to put back to the pool the points used for blitting.

### DIFF
--- a/flixel/addons/display/FlxBackdrop.hx
+++ b/flixel/addons/display/FlxBackdrop.hx
@@ -49,10 +49,9 @@ class FlxBackdrop extends FlxSprite
 	 */
 	public var blitMode:BackdropBlitMode = AUTO;
 	
-	var _tileSize:FlxPoint = FlxPoint.get();
-	var _tileMatrix:FlxMatrix = new FlxMatrix();
 	var _blitOffset:FlxPoint = FlxPoint.get();
 	var _blitGraphic:FlxGraphic = null;
+	var _tileMatrix:FlxMatrix = new FlxMatrix();
 	var _prevDrawParams:BackdropDrawParams =
 	{
 		graphicKey:null,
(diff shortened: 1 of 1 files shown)


Unholywanderer:
mid pr 🥱
would never get merged

Geokureli:
I also found a leak in backdrop so i may add it here if you didn't already

can I get simple steps to reproduce the leak?

MaybeMaru:
> I also found a leak in backdrop so i may add it here if you didn't already
> 
> can I get simple steps to reproduce the leak(s)?

FlxMatrix is a pretty cheap class to create so you have to get to some huge numbers of instances of FlxBackdrops for it to affect on memory much, but still its slightly better to have the matrix cached rather than creating new instances.
Not sure if thats the leak youre referring to as well though

Geokureli:
I think you missed this part:
> > can I get simple steps to reproduce the leak(s)?

flixel/addons/display/FlxBackdrop.hx
```diff
@@ -311,57 +316,54 @@ class FlxBackdrop extends FlxSprite
 		getScreenPosition(_point, camera).subtractPoint(offset);
 		var tilesX = 1;
 		var tilesY = 1;
-		if (repeatAxes != NONE)
```
Geokureli, line 314: why was this taken out?

Geokureli: reverted this

Geokureli:
I like to avoid instance points/rects in favor of local function get/put calls, for many reasons, this was the change I had made last week: https://github.com/HaxeFlixel/flixel-addons/pull/427

FWIW: after profiling I found that matrix instances were a tiny part of garbage collection in my tests, a majority of these instances were from openfl classes and we have very little control over those. but nonetheless, this should still help

MaybeMaru:
aight looks good

PR #441, opened 2024-07-14
Remove deprecation warning on Flixel 5.9.0

richTrash21:
Makes FlxTransitionSprite use onFinish instead of finishCallback on Flixel 5.9.0

Caused by https://github.com/HaxeFlixel/flixel/pull/3205


Geokureli:
Thanks!

PR #444, opened 2024-09-24
Refactor FlxPieDial, add FlxRadialGauge

### LINKED ISSUE #430: FlxPieDial black border
Upon creating a FlxPieDial, it renders with a black border around it. That border disappears as the pie empties out. In the HaxeFlixel demo of it, the background is black itself, so it's not noticeable. Might have been an uncaught bug or something breaking as time passed on.

example:
image

Geokureli:
## FlxPieDial
- Fixed issue where the background of the dial was black, instead of transparent. Fixes #430
- Moved drawing logic to util, so it can be used by `FlxRadialGauge`

## FlxRadialWipeShader
- Shader that masks sprite radially, can be applied to of any static sprite (doesn't work with animations, yet, but neither did FlxPieDial). Shaders only work on non-Flash targets

## FlxRadialGauge
Meant to replace `FlxPieDial` on non-flash targets, better on all fronts.
- Uses `FlxRadialWipeShader` via `setOrientation(start, end)`
- `FlxPieDialUtils` to allow you to draw a radial gauge akin to `FlxPieDial` via `makeShapeGraphic(shape, radius, innerRadius, color)`

https://github.com/user-attachments/assets/9446ef42-b526-4fb1-a59f-909273769329

### DIFF
--- a/flixel/addons/display/FlxPieDial.hx
+++ b/flixel/addons/display/FlxPieDial.hx
@@ -1,18 +1,23 @@
 package flixel.addons.display;
 
 import flixel.FlxSprite;
+import flixel.graphics.FlxGraphic;
+import flixel.graphics.tile.FlxGraphicsShader;
 import flixel.math.FlxMath;
 import flixel.math.FlxPoint;
+import flixel.util.FlxBitmapDataPool;
 import flixel.util.FlxColor;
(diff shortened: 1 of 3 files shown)


Vortex2Oblivion:
Would be a cool addition!

PR #449, opened 2024-10-07
fix: use `parallaxx` and `parallaxy` from the tiled tmx

ninjamuffin99:
in Tiled 1.5 `parallax` related variables were added to layers. This PR will read the layer attributes from the tmx file and set the `parallaxX` and `parallaxY` values if it exists in the tmx, otherwise defaults to 1


Geokureli:
Thanks!

PR #452, opened 2024-12-21
add flixel as a dependency

Vortex2Oblivion:
lime and openfl were added to the base flixel as dependencies so it would make sense to add normal flixel to flixel addons as a dependency


Geokureli:
Thanks!

PR #454, opened 2025-01-23
check correct lib versions

Geokureli:
Note that in compiler conditionals, `#if ("5.10.0" < "5.9.0")` is true, however `#if (version("5.10.0") < version("5.9.0"))` is false.


moxie-coder:
I didn’t know you could do that

PR #462, opened 2025-06-16
FlxGridOverlay: Prevent freeze when passing invalid cell size values

ACrazyTown:
Passing in 0 or a negative number (in my case due to a bad calculation) when creating a grid overlay caused the while loops in `createGrid()` to never exit because x/y would always be smaller than width/height

https://github.com/HaxeFlixel/flixel-addons/blob/cff2640a6db35544eb3faf3655de5e57a709cc2a/flixel/addons/display/FlxGridOverlay.hx#L112

### DIFF
--- a/flixel/addons/display/FlxGridOverlay.hx
+++ b/flixel/addons/display/FlxGridOverlay.hx
@@ -44,10 +44,11 @@ class FlxGridOverlay
 			Height = FlxG.height;
 		}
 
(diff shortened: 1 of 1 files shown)


Geokureli:
I'm not familiar with this class, is there an example usage somewhere?

Also, does it make sense to throw or log an error in this case?

Geokureli:
Also, I don't recommend using this class, lol. this seems outdated and optimized for Flash rather than renderTile targets. I'm still okay with the PR, but yeah I think this class should be replaced and deprecated

PR #465, opened 2025-09-03
FlxNapeSpace: properly handle and destroy plugin instance

ACrazyTown:
Currently FlxNapeSpace has an issue where calling `FlxNapeSpace.init()` multiple times (eg. due to a state reset) would mess up physics because there'd be multiple active plugins updating the space in a single frame.

I've added an `instance` variable that tracks the created plugin and destroys it when trying to init() again

### DIFF
--- a/flixel/addons/nape/FlxNapeSpace.hx
+++ b/flixel/addons/nape/FlxNapeSpace.hx
@@ -23,6 +23,11 @@ private class GraphicNapeDebug extends BitmapData {}
  */
 class FlxNapeSpace extends FlxBasic
 {
+	/**
+	 * The instance of the global FlxNapeSpace plugin, created when calling init()
+	 */
+	public static var instance:FlxNapeSpace;
+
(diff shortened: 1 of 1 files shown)


Geokureli:
There's potential for different desired behaviors, here:
1. They want the second call to wipe the first (yours)
2. They want the second call to be ignored
3. They want two instances

This prohibits options 2 and 3, I say we force them to check for an existing instance, or clean up on state change, or whatever they want to happen

Geokureli:
If having two instances doesn't work in any real way then yeah I would throw an error on the second call,

> It's also really inconvenient to work with. If you call `FlxNapeSpace.init()` at the start of your state (which is as far as I know where you're supposed to put it) you can't even use the debugger's reset feature because it'll break physics.

I tend to init cross-state managers and plugins in either Main.hx, some BootState that only gets run once, or on `FlxG.signals.preGameStart.add(()->initStuff());`

PR #475, opened 2025-12-20
Continuation of Flixel PR #3538

rodney528:
https://github.com/HaxeFlixel/flixel/pull/3538


Geokureli:
Exists is set to false in destroy, anyway

PR #476, opened 2026-01-19
Add default values to FlxTextField constructor

Vortex2Oblivion:
Updated constructor parameters to have default values for X, Y, and Width.

### DIFF
--- a/flixel/addons/text/FlxTextField.hx
+++ b/flixel/addons/text/FlxTextField.hx
@@ -25,7 +25,7 @@ class FlxTextField extends FlxText
 	 * @param	Size			The font size for this text object.
 	 * @param	EmbeddedFont	Whether this text field uses embedded fonts or not
 	 */
-	public function new(X:Float = 0, Y:Float = 0, Width:Int = 0, ?Text:String, Size:Int = 8, EmbeddedFont:Bool = truee)
+	public function new(X:Float = 0, Y:Float = 0, Width:Int = 0, ?Text:String, Size:Int = 8, EmbeddedFont:Bool = true)
 	{
 		super(X, Y, Width, Text, Size, EmbeddedFont);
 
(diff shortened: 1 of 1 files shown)


flixel/addons/text/FlxTextField.hx
```diff
@@ -25,7 +25,7 @@ class FlxTextField extends FlxText
 	 * @param	Size			The font size for this text object.
 	 * @param	EmbeddedFont	Whether this text field uses embedded fonts or not
 	 */
-	public function new(X:Float, Y:Float, Width:Int, ?Text:String, Size:Int = 8, EmbeddedFont:Bool = true)
+	public function new(X:Float = 0, Y:Float = 0, Width:Int = 0, ?Text:String, Size:Int = 8, EmbeddedFont:Bool = truee)
```
dtwotwo: please fix - `truee`

moxie-coder: lmao

Vortex2Oblivion: oops lol

Geokureli:
Are you using this class is a project?

Vortex2Oblivion:
> Are you using this class is a project?

I was thinking about maybe trying it but I just noticed that it didn't have default constructor values while the superclass did and thought it was a bit strange

Geokureli:
I was thinking about deprecating this class, entirely. I don't think it has any uses, now

Vortex2Oblivion:
Still, it is kinda strange that there aren't any default values

Geokureli:
Not as strange as maintaining a class no one uses. I'm always afraid that updating these classes will make people think they're worth using. I might just deprecate all these old classes and see if anyone complains. If you do end up using it, let me know, since it tells me there's an issue with using FlxText

PR #478, opened 2026-02-28
FlxTypeText: Fix FlxSound deprecation warnings

KELOTOMU:
Fix deprecation warnings in favor of FlxSound reworked load names


Geokureli:
These need to be wrapped in `#if (flixel < version("6.2.0"))` checks for backwards compatibility

PR #483, opened 2026-05-07
Fix an issue where hitboxes didn't work on FlxEffectSprite

ADA-Funni:


flixel/addons/effects/chainable/FlxEffectSprite.hx
```diff
@@ -132,9 +132,18 @@ class FlxEffectSprite extends FlxSprite
 			_flashRect = pixels.rect;
 		}
 
+		updateHitbox();
+
 		super.draw();
 	}
 
+	override public function updateHitbox() {
```
Geokureli: Don't cuddle brackets

Geokureli (changes requested):
All PRs should include a small code example that reproduces the issue, and explains the expected behavior vs the actual behavior

Also you need to match the flixel's code style, curly brackets go on a new line

flixel/addons/effects/chainable/FlxEffectSprite.hx
ADA-Funni: It's fixed now, sorry.
I also defined it as `Void` because all the other functions do that

ADA-Funni:
> All PRs should include a small code example that reproduces the issue, and explains the expected behavior vs the actual behavior

Here's a couple diagrams of the behaviour

Before my PR:
image

After my PR:
image

ADA-Funni:
The new behaviour (shown below) makes the calculations from `updateHitbox()` work in terms of the target's height/width, whereas the original used the effect sprite's height/width, which got really weird due to the hitbox code staying untouched.
image

The old behaviour was just the regular `updateHitbox()` function applying to the effect sprite

Geokureli:
Could we solve this by setting frameWidth and frameHeight to the target's frame sizes? I'm pretty sure there are other methods that use this.frameWidth and will also need updating

ADA-Funni:
> Could we solve this by setting frameWidth and frameHeight to the target's frame sizes? I'm pretty sure there are other methods that use this.frameWidth and will also need updating

I think that could work?
I'll try that later, thank you. :D

